PR #7, opened 2017-02-05
portability changes for building

dorkulon:
I've mucked about with the Makefile and the python script for generating the file listing for start.tw; it successfully builds the start.tw file as expected on both Cygwin (on Windows) and on BSD. It does explicitly use GNU make-isms, but I don't think that's a big deal at this point.

If start.tw could use ::@include for the file list, that would be even better - but alas, that does not seem to work.


ObstacleCorpse:
Thank you very much for taking the time to do this. Unfortunately I don't have access to a GNU/Linux system right now, but I'll toss up a VM in the morning and verify that these changes work as intended. If so, I'll merge the request in. Thanks again.

dorkulon:
One thing I found during testing: tweego.exe does not care about the path separator! It can deal with either src\foo\bar.tw or src/foo/bar.tw without issue. Similarly, it does not care if the line endings are the Windows \r\n or the Unix-y \n; this eliminates an awful lot of the tomfoolery required with python and line endings.

Note that these changes break the Windows batch script! I have an additional commit that changes it to use the includes_linux script instead: https://github.com/dorkulon/Free-Cities/commit/1c36de5580d7373bc3632b08c69584dba339bad2

ObstacleCorpse:
Tested on Fedora 25 x86_64, and everything works as expected. Fantastic job; thanks again.

PR #10, opened 2017-02-06
change to rules assistant.tw

for-anonimity:
Now it correctly displays when slaves are excluded or only apply to some slaves.
It didn't work because "and" has an higher priority than "or"


ObstacleCorpse:
Fantastic, thanks. Flagged for FC Dev's review.

freecitiesdev:
Merged, thanks.

PR #26, opened 2017-02-09
RA: hijab and abaya

wepsrd:
Forgot about the rules assistant!


freecitiesdev:
I, too, forget about the rules assistant. All the damn time.

PR #28, opened 2017-02-10
Bugfixes №4

Folant:


src/uncategorized/slaveInteract.tw
```diff
@@ -163,7 +163,7 @@
 			<</if>>
 		<</if>>
 	<</if>>
-	<<if ($activeSlave.rivalry != 0)>>
+	<<if ($activeSlave.rivalryTarget != 0)>>
```
Folant: Should fix a situation when there's no rival but the button is active, resulting in two errors. 
rival2

PR #29, opened 2017-02-10
Implement adding all the tw files to StoryIncludes in the shell, so no need for Python or Make

blit9:
This code does the same task that devTools/scripts/includes.py performs, but does it directly in the shell on both Windows and Linux/OSX etc.

The changed compile script works for me, but open to reworking it if it's got any issues.

Since that's the only thing that Python/Make is being used for in the project, I removed those as well.


freecitiesdev:
Will this require updating StoryIncludes when new .tw files are added?

ObstacleCorpse:
@freecitiesdev 
No. I have only looked at the commit, and have not tested it yet, but it should replicate all of the functionality of the current compiler setup (including adding new .tw files).

Part of the reason a Python-based solution was originally chosen is that, in other projects, we've made more extensive use of JavaScript, and had files that needed to be loaded in a specific order. Python scripts were used for this, and the Makefile allowed for various types of compilations, depending on what we were testing. However, none of that currently applies to FC. So, unless you expect to heavily expand this project, I see no issue with the proposed changes.

blit9:
@freecitiesdev 
@ObstacleCorpse 

Sorry for the late reply.

I believe what ObstacleCourse said is correct.

I've got nothing against Python or anything - if you are intending on expanding the role of Python in the projet I'd say keep it; otherwise if you're only using Python to generate the StoryIncludes list, then I think Python might be a bit of overkill. My change does the same job the Python script is doing, but like you say it doesn't maintain a specified order of includes.  I don't think that matters for FC, like you say.

Same for the Makefile really - if you are going to use it extensively I think it's worthwhile, otherwise it's a bit of overkill just to run a couple of commands.

blit9:
Oh another note - my changes should work on both Windows and Unixes (Linux, OSX etc). I have tested on Windows and Ubuntu, but I don't have a Mac handy.

PR #30, opened 2017-02-10
Update policies.tw

DrNoOne0:
Should be a fix to the active Future Society Policies section of Policies not showing up.


DrNoOne0:
Bug fix doesn't work.  Investigating further.

DrNoOne0:
Should be good to go.  Problem was caused by an incorrect variable name during init.  Fix includes a temporary solution in policies.tw that should stay until mod_BackwardsCompatibility.tw gets updated.

freecitiesdev:
What was the bad variable in init?

DrNoOne0:
The bad variable init was $arcologies[0].FSGenderRadicalistLawBimbo.  I changed to $arcologies[0].FSGenderFundamentalistLawBimbo, since that is what the policies.tw passage was looking for.

PR #96, opened 2017-03-09
Bugfix saRelationships

Bane70:
Fixes Family Feelings section of code from not firing for some needed slaves, cleaned up a few unneeded _vars in a part of the code, and stopped saving the slaves if they are a fuckdoll.


Bane70:
Arg. It's sending the wrong updated files.

PR #110, opened 2017-03-12
Bugfixes-Optimizations dairyReport.tw

Bane70:
Fixes issues with the $Milkmaid var and her $slaves[] var being out of sync. Also used canPenetrate() for her. An I fixed her penetration counts to be the same # of times as for a slave (+10 for each)
Changed Line 420 <<if ($Milkmaid.muscles > 1)>> to be <<if ($Milkmaid.muscles > 30)>> As that seems to be a left over from the old system.

Made some of variables as _vars since they are only used in this passage _anusesStretched, _birthers, _births, _cumWeek, _femCumWeek, _milkmaidFetish, _milkmaidImpregnated, _milkWeek. The _birthers, _births vars don't seem to do anything anymore (beyond disabling the related texts for them), but I left them in just in case it's for future use or something.

### DIFF
--- a/src/uncategorized/dairyReport.tw
+++ b/src/uncategorized/dairyReport.tw
@@ -2,259 +2,220 @@
 
 ''Dairy Report''<hr style="margin:0">
 
-<<set $dairySlaves to 0>>
-<<set $milkWeek to 0>>
-<<set $cumWeek to 0>>
-<<set $femCumWeek to 0>>
-<<set $birthers to 0>>
-<<set $births to 0>>
-<<set $anusesStretched to 0>>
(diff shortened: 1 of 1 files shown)


000-250-006:
@Bane70 

I get a bug from this (I think line 4) -  returns

> Error: <<set>>: bad evaluation: assignment to undeclared variable drop

Folant:
@000-250-006 https://github.com/Free-Cities/Free-Cities/pull/108/files?diff=split ?

000-250-006:
Annnnd this would be me whistling casually as I slide stage left innocuously :P

PR #113, opened 2017-03-12
Bugfix-Opt servantsQuartersReport.tw

Bane70:
Fixes issues with the $Stewardess var and her $slaves[] var being out of sync. Also used canPenetrate() for her. An I fixed her penetration counts to be the same # of times as for a slave (+10 for each). Also separated her code from the other workers, and fixed a possible bug with her impregnating counts not reflecting all possible workers impregnated.

Again var were changed over to _temps; _SQMilk, _SQMilkSale, _stewardessFetish, _stewardessImpregnated.

Not sure if this meant to be this way or is an oversight due to the $universalRulesFacilityWork, but "SA servant" already does  <<set $cash += $stewardessBonus>> for up to a $200 bonus, and then it will also do it again here on **line 115** for another bonus. **I left it in** just in case it's on purpose for SQ workers to get up to $400 bonus.

### DIFF
--- a/src/uncategorized/servantsQuartersReport.tw
+++ b/src/uncategorized/servantsQuartersReport.tw
@@ -2,14 +2,53 @@
 
 ''Servants' Quarters Report''<hr style="margin:0">
 
-<<set $servantsQuartersSlaves to 0>>
-<<set $stewardessFetish to 0>>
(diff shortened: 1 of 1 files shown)


freecitiesdev:
That's an oversight; I'll get it.

PR #124, opened 2017-03-14
accordion hotfix

000-250-006:
Somewhere along the line today Sugarcube decided to be a lot more picky about using ndef as a flag, broke this. Fixed it (although I hate having to take up variable space to do it)

### DIFF
--- a/src/js/accordionJS.tw
+++ b/src/js/accordionJS.tw
@@ -15,11 +15,11 @@
 */
 
 postdisplay["doAccordionSet"] = function (content) {
-    if (variables().useAccordion == 1) {
-        Array.prototype.slice.call(document.querySelectorAll('.macro-display'))
-        .forEach(function(element) {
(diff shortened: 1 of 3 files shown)


Bane70:
You may need to declare your variable now somewhere since it's not using the def and ndef. If someone doesn't go to options to set it it'll throw an error at "Slave Assignments Report" of undefined variable. lol

000-250-006:
It should be OK, I still had the old init file in there from when I'd forgotten to add the story initialization toggle.

I just wish I could figure out what changed to break it. 3 days ago it was fine. The repo conflicts were purely cosmetic text formatting, so that wasn't it.

Mondays, amirite?

Bane70:
OK. Ah that reminds me there was a var I had changed to a _temp one that's in the storyinit. Lol now to figure out which it was.

It seemed to work fine for me when I compiled it a couple of hours before this update.

PR #131, opened 2017-03-14
Click/display macros deprecation

Folant:
Since  click macro is deprecated in  SugarCube 2.8.0 and display is deprecated in 2.15.0 should we replace them wherever possible? If not, just close this PR.


freecitiesdev:
Have you tested this?

When the click macro deprecation was first mentioned, I ran into trouble replacing the existing incidences, but I don't recall what the exact issue was. If you've tried this and it works, though, objection withdrawn.

Folant:
Yes. I ran two versions at the same time, and everything seems to work fine without errors. Without this, the code already had 18 instances of the link macro, as shown by  ```git grep '<<link' | wc -l```

freecitiesdev:
Outstanding. Forgive the automatic skepticism. I wish I could add a "have you run this to check for glaringly obvious error messages?" confirm box for pull requests.

freecitiesdev:
It does seem to have broken the keyboard shortcuts, though.

Folant:
Hm. I will take a look now, but they seem to be broken even without it.

freecitiesdev:
Interesting. I was doing (unrelated) playtesting right before pulling that code, and they were working fine for me.

Bane70:
Were any of those clicks referring back to the mouse routines in JavaScript? I had once edited some clicks before and killed the hotkeys.

Folant:
@freecitiesdev It seems storyCaption.tw click --> link really broken shortcuts, damn.

Folant:
@Bane70 yeah, seems so. There are 16 click macros in storyCaption.tw 
And all of them are refered in userlib.js like that, I believe
```Mousetrap.bind("enter", function () {
	$("#story-caption #endWeekButton a.macro-click").trigger("click");
});
Mousetrap.bind("space", function () {
	$("#story-caption #nextButton a.macro-click").trigger("click");
});
Mousetrap.bind("c", function () {
	$("#story-caption #manageArcology a.macro-click").trigger("click");
});```

Folant:
Maybe if I change trigger("click") to trigger("link") it'll work.

Bane70:
Maybe. I'm no JS expert though.

Bane70:
Lol come to think of it that was the same passage I had edited when I killed the hotkeys, that or costs.

Folant:
Fixed in #132

PR #137, opened 2017-03-15
Update cellblockReport.tw

Bane70:
Just bugfixed. I'll opt it when I have more time as I gotta find something to eat before midnight, as I'm having a surgery for a cataract done in the morning. Both the spa and this one has the if married to PC issue but I figured out how to display that I'll work on it later tonight or maybe tomorrow depending on how I feel.


freecitiesdev:
Best of luck.

PR #141, opened 2017-03-16
SVG Fix

000-250-006:
As requested ...

For the record, whoever decided to have Sugarcube check tag completion within conditionals is a ███████ ██████ of ████ that should be █████ by the ███████ neck until they ████ █████████ in terror!*

*NSFW Censored

### DIFF
--- a/src/gui/css/mainStyleSheet.tw
+++ b/src/gui/css/mainStyleSheet.tw
@@ -11,33 +11,38 @@
     max-width: 76em;
 }
 
-div.imageRef{
-margin-left: -75px;
-margin-right: -75px;
-overflow:hidden;
-width: 400px;
-height: 400px;
-position: relative;
-float:right;
+.imageRef {
+    display: flex;
+    flex-direction: column;
+    flex-wrap: wrap;
+    align-items: flex-start;
+    position: relative;
 }
 
-div.lowPenis{
-width: 400px;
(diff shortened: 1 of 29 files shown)


freecitiesdev:
This leaves space for images in lists when images in lists are disabled.

Art enabled + images in lists disabled = spaces for nonexistent images in lists.

Bane70:
Lines 14 and 51 used $i instead of _i.

freecitiesdev:
In which file?

freecitiesdev:
Never mind, I see them. That's where he's commented the old code out.

000-250-006:
OK, catching up.

I didn't think about the blanks in lists, my bad. I'll work that one on the morrow.

The cropping part is easy in theory - the parent div using flex, combined with object-fit, gives it a lot of flexibility.  It's a lot easier, and has less chance of having odd aspect ratios blow up in your face, than using neg margins.  The problem in this case is going to be making sure it stays consistent across both image formats - render and SVG.

The SVG images are 1000x1000, rendered are something like 531x531. Rendered images use more of their horizontal space than the widest SVG does, so it'll take some math and tweaking. Just time and testing, not a biggie.  

@Bane70 I'll check for the variables you mentioned. It's entirely possible I borked redoing them across. That's why I leave originals in the code until it's good :)

PR #145, opened 2017-03-16
Club report fix

Folant:


Bane70:
I don't think the DJ is supposed to get bonuses for each of the Public service workers, just the once if she has to also work the streets.

Folant:
Well, without that I get these messages if there are less than 5 slaves. Any advices?
i

Bane70:
Hmm that's odd since "SA serve the public" handles the beauty widget. I just had to place it here so the summary text would represent those vars right then it zero's them out before the slaves actually run "SA serve the public". An I've not had that error as of yet. I'll dig into the widgets. Looks to be in <<Modscore>> or I didn't set something right somewhere. Maybe a widget is set to only use slaves[$i] or something, but that should work since I set $i to = _i or whatever mt temp var is at the time for the slaves. Or I should be at least.

Folant:
🍀 Thank God commits can be deleted, lol

Bane70:
Lol. I'm working on a Macro (Macro.add) that would work much like << silenty >> but save the text to a variable so that it could be displayed later. All that would have to change for the end of week reports is to use the new macro instead of << silently >> then check after the summary for <<if $showEWD != 0>> and have it output the variable text and probably then set the var to "" as it can be quite big.

Folant:
Aaand it fixes that, but breaks gazillion of other things 😁

Bane70:
Lol hopefully not. Right now I'm literally using the silently code, and just trying to get it to place things into my variable but it keeps giving me the argument is read only. But that was last night and I might of been still under the influence of the drugs the hospitol gave me for my surgery :)

Folant:
Well this is the best solution I can offer now, since no other is present atm, lol. Maybe @freecitiesdev can lend a few words.

Bane70:
I could be wrong but I think << Beauty $Milkmaid (or whatever slave)>> already did accept slaves[$i] or whatever as a variable or at least it appears that way from the "SA public service" (since I copied directly from and pasted in as needed.) and so does <<modScore $args[0]>> as seen by it having $args[0], that should be the slaveObject.
I've still not seen your errors pop up for me as of yet 

There is a missing <<set $i = _i>> in clubRerport I just noticed inset it just after the if on line 132

freecitiesdev:
<<Beauty $Milkmaid>> should work.

Bane70:
The only reason you might get an args[0] is undefined is if somehow the code thinks you have a Dj when you actually might not.

Folant:
Well, <<Beauty $X>> also works, so I leave it that way.

Bane70:
That's probably the best way to call beauty since it's using a slave directly, and not having to search $slaves[_DJ] in the array. Though it bugs me as to why the other beauty above it works and not the _DJ one since _DJ is just the $i/_i index number of the slave.

Folant:
@Bane70 
> what was those bad conditions in the errors you got?

For the science: 
https://gist.githubusercontent.com/Folant/299e3036044e5b6d6e4a385f4df0cdd6/raw/c475f56a1720f6c2e6a156528a776713aaac0d12/error

Bane70:
Hmm going by that it kinda looks like an out of bounds slave maybe. >= _SL. But if using $DJ works then that's not the case.

Bane70:
It bugs me not because I might of made a mistake somewhere but that it might mean all the code in lines 27 to 55 would also be wrong and also should be undefined since they also use $slaves[_DJ], and line 55 even sets $DJ = $slaves[_DJ]. An I'm at a loss as to what might be the error.

Folant:
Take a look at my save, if you want. I can reproduce it since Week 1 with completely random settings. 
[free-cities-20170317-013748.zip](https://github.com/Free-Cities/Free-Cities/files/849263/free-cities-20170317-013748.zip)

Bane70:
I think I might just to make sure. Cause if I coded something wrong then all the facilities will probably have this issue. Since they are all coded more or less the same to get their leaders index #.

Thanks.

Bane70:
Did you have a DJ before you ended the week? as I completed the week and there is no DJ appointed.

Folant:
>Did you have a DJ before you ended the week?
No.

Folant:
If I appoint DJ - the errors disappear

Bane70:
Then that explains it ... I assumed that you had one. Oh man my bad now I see my error. there should be an <<if $DJ != 0>> or <<if _DJ > -1>> after the <</for>> on line 136 and the closing if after line 150.

Folant:
Good find 🥇

Bane70:
Lol too bad it only took 18+ hours.

Folant:
More, since I've been trying to figure out the problem since yesterday -_-

Folant:
Anyway, do we actually need to add it now?

Bane70:
Ouch. Hmm I'm guessing that the $DJ var works as all it'll give is 0's to things (or if like C language garbage values) like $DJ.anus etc. when it should also be erroring out since $DJ = 0, if it was $DJ = null it might of errored out as well, instead of giving a false impression of it working. null is considered an object, and 0 is a number

Bane70:
That depends on the way JavaScript handles things is it just silently skipping things or is it using garbage values? It would be a bit faster (probably unnoticible unless you have hunderds of slaves) to skip running the code if it's not needed to run

Folant:
¯\_(ツ)_/¯

Bane70:
I can't find the document I once read but if I remember right JS would be looking for the .anus or whatever in the object via text string search. So I'm guessing that it's silently skipping things, or passing a 0 to things. But when $DJ = 0 it should be erroring out as well since it's a number and can't contain .anus a part of an object, or so it would do in any other language. lol.

PR #148, opened 2017-03-16
Milkmaid Bonus Fixes

DrNoOne0:
This fixes the incorrect Milkmaid bonuses. They likely have been wrong since the change to integer devotion and trust.

Change list:
- Fixes the Milkmaid's devotion, trust, and health bonuses provided to cows.
- Makes the Milkmaid build muscles on the job to be better at her job.
- Makes the Milkmaid learn oral sex skills on the job, also to be better at her job
- Changes the Milkmaid's health bonus to be curatives.
- Changes the Milkmaid's special attention code to be (hopefully) slightly faster.  Likely not noticeable.


Bane70:
The deletion of the << set $i = _i >> on line 97 of the orig dairyReport is needed to work with the << includes >> for the slaves otherwise it is probably just recalculating the MilkMaid over and over again.

DrNoOne0:
I thought I had left that in...  It is now back, thank you for catching that.

PR #156, opened 2017-03-18
HOTFIX Accordion Update

000-250-006:
Fixed a couple of bugs

Redid CSS display


000-250-006:
I still need to poke some more to detect if the special slaves in facilities are present alone, and to get their names for the display, but I wanted to get back on the SlaveArt widget today so, this one is serviceable (I hope)

000-250-006:
Closing this, final ver in testing.

PR #159, opened 2017-03-19
Reformatted Facility Texts/Images fix sorta.

Bane70:
Fixed a few minor bugs and hopefully got a decent reformatted text now. I got rid of the blank squares of where the vector art would appear by checking for <<if $seeImages == 1 && $imageChoice == 0>>. Looks good and the same now no matter what the options are art/no art, and with or without the accordion.
Hopefully the next five won't take me as long.

NOTE: I had to delete some code that I'm sure was supposed to be a fix but it wrecked havok with trying to get the text right for all displays. But it works fine where it displays the slave information/description.

If image art is turned on it causes an extra line break I'm assuming because of this <<if $seeImages == 1>>< br style="clear:both" /><</if>> I left it in as I'm not sure as to what it's supposed to do as to what's clearing.


000-250-006:
I'm going to be revisiting that work-around once Accordion is out the door, I'm just waiting to hear if I can chunk up the SVG images (copyright)

PR #160, opened 2017-03-19
dSlavesDatabase Fixes

Bane70:
Fixed some missing _HS.'s


Bane70:
I think I finally got them all.

PR #161, opened 2017-03-19
Accordion Final

000-250-006:
Nearly final, really. Have some investigation to do about best memory practices with it.

But, this is RC down code. Comments inline.

PR #162, opened 2017-03-20
endWeek.tw fix

Folant:
https://github.com/Free-Cities/Free-Cities/pull/161#issuecomment-287657966

### DIFF
--- a/src/uncategorized/endWeek.tw
+++ b/src/uncategorized/endWeek.tw
@@ -73,7 +73,7 @@
 	<<set $PC.sexualEnergy += 2>>
 <</if>>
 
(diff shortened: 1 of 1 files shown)


000-250-006:
OK, I've been tweaking Accordion too much and it's starting to look like a mess in Git, so I'm freezing it at the current release (which is stable) and giving @freecitiesdev a chance to catch up, review, see if they want it, and the like.

Folant:
Good work, @000-250-006. lol.
git 1

000-250-006:
Xkcd ... damn, and here was me thinking I'd come up with that solution all by my lonesome :P

PR #169, opened 2017-03-22
Dairy as an Array

Bane70:
Actually in my game I've had no noticeable problems but I currently only have 8 slaves that can use the dairy, so I can't say for sure if it's actually going faster. Main menu does seem to be a bit better but that's from the changes to slaveSummary this morning I think as it was better before I added the dairy changes.
Also of note this requires those changes to work right so the slaveSummary.tw update is required for this to work right.
If there are no major issues with this there are still area's of improvement like I know there's a custom event that might be able to use the array directly instead of searching the slaves[..] array.

The $DairyiIDs array only contains the .ID and .Index The .ID is the slaves[..].ID number, and the .Index is the [..] number in the slaves array. It check at the beginning of loops to make sure it has the right .ID slave before processing things. If it don't it will search the slaves[..] for the .ID of the slave and then update $DairyiIDs.Index. Due to the player sorting his slaves or maybe an error I made.

This is more or less just a test release as I have not added it yet into startinit, or NG+. I just placed a few if ndef's for now. so that I didn't have to start a new game to test it out.

I also gotta search if there are more aea's that remove the slave from the job and doesn't use Retrieve to remove her. So expedct additions in a bit.


Bane70:
Those conflicts should be ok as I used the same files minus the Dairy stuff in them, so a total overwrite would work.
It seems there is no other area's the slaves are removed from the dairy, or deleted from the game that I can find. so this should be good.

Going by the email, all that is required in startinit is a set for $DairyiIDs = []

Issue #168, opened 2017-03-21
SugarCube


Bane70:
Just noticed in the Sugarcube documentation it says that < Array>.contains(), < Array>.containsAll(), and < Array>.containsAny() has been deprecated: since v2.10.0. and should start using the < Array>.includes equivalents.
From what I can see < Array>.contains() seems to be still working right for the Careers.

Folant:
> From what I can see < Array>.contains() seems to be still working right for the Careers.

It's true for every deprecated thing, I think. Like in #131 .
`cd src && git grep -c '\.contains' * | wc -l && git grep '\.contains' | wc -l` 
gives 134 occurences in 36 files, so I guess I just run 
`git grep -lz '\.contains' | xargs -0 perl -i'' -pE "s/\.contains/\.includes/g"` 
and create PR when current changes are merged.

PR #183, opened 2017-03-29
SugarCube contains macro deprecation

Folant:
#168


freecitiesdev:
Thanks. Haven't been keeping up on all these deprecations lately.

PR #188, opened 2017-03-30
Spa report fix

Folant:


Bane70:
Ah I see why it's not working now. lol. Wonder when I messed up the Index/ID. Since it did work fine at one point. My version of Sublime might have issues with searching and replacing it seems, since it missed one (wii). Thanks!

PR #189, opened 2017-03-31
fix toychest bug, remove left-over events

goldifrocks:
toychest is occassionally referencing $activeSlave, which may lead to interesting results.

the 'RE shift' events are leftovers and are now duplicates of the similarly named 'shift' events.


freecitiesdev:
Incorrect $activeSlave references are a key thing to watch for, given the consequences of the way I've set this $slaves nonsense up. Thanks.

PR #190, opened 2017-03-31
Update slaveSummary.tw

DrNoOne0:
This update does three things:

1. Prevents slaves from being listed for assignment to the facility they are already in.
2. Uses break when the facility is full, eliminating a pointless cycling through the slave list in such a scenario.
3. Gives a simple message stating that the facility is full.

### DIFF
--- a/src/uncategorized/brothel.tw
+++ b/src/uncategorized/brothel.tw
@@ -147,7 +147,9 @@ $brothelNameCaps
 	<<include "Slave Summary">>
 <</if>>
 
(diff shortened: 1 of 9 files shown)


Bane70:
Nice! I just noticed though that the statements with the breaks would be better off in each of their facility files instead. That way it would not even start to run the loop, and eliminate a bit of code that gets needlessly continuously tested otherwise.

DrNoOne0:
That makes too much sense.  Making that change now.

Issue #170, opened 2017-03-22
Missing pull?


000-250-006:
#161 was closed but not merged. There's 3 commits in there essential.

000-250-006:
Patch file against upstream/master
[accordion.zip](https://github.com/Free-Cities/Free-Cities/files/862466/accordion.zip)

Folant:
https://github.com/Free-Cities/Free-Cities/issues/171#issuecomment-288346649 : 
@000-250-006 
https://github.com/000-250-006/Free-Cities/blob/8b8bdbcd4d3dcf0edd42915f4096138339c8cb43/src/uncategorized/slaveAssignmentsReport.tw#L317
should be `["Spa Report", $spaNameCaps, $spa, $spaSlaves, $Attendant, "Attendant"]` .

000-250-006:
Confused ... that's what it is already?

`["Spa Report", $spa, $spaNameCaps, $spaSlaves, $Attendant, "Attendant"],`

000-250-006:
This is me banging my head against the desk

Good catch, thanks!

Bane70:
Is the patch file still needed? As I seem to have in options the toggle is backwards, disabled is actually enabled. and at least the spa isn't appearing. An I already switched around ["Spa Report", $spaNameCaps, $spa, $spaSlaves, $Attendant, "Attendant"].

Folant:
>Is the patch file still needed? 

Well, I guess it doesn't contain https://github.com/Free-Cities/Free-Cities/issues/170#issuecomment-288519533 at least. #187 also changed slaveAssignmentsReport.tw a bit.
I don't know if @freecitiesdev is going to reopen and then merge #161, or @000-250-006 is going to open new PR later ( if so, when? ). Here's the latest correct version with all of the above changes though.
[accordionfinal.zip](https://github.com/Free-Cities/Free-Cities/files/884112/accordionfinal.zip)

Bane70:
Ah I see what I did, I capitalized $spa for some reason in ["Spa Report", $spaNameCaps, $spa, $spaSlaves, $Attendant, "Attendant"]. Lol that doesn't help anything.

Folant:
>Lol that doesn't help anything.
>As I seem to have in options the toggle is backwards, disabled is actually enabled. 

I can't understand the problem, maybe because it's nearly 4 AM here lol

Bane70:
[...]

PR #191, opened 2017-03-31
Fix HeadGirls Girl slaveAssignmentsReport.tw

Bane70:
I totally forgot to test out the head girls girl. lol. Also corrected the wrong placement of $spa for the accordion.


Folant:
Lol, I wonder, maybe it's better to incorporate all other Accordion changes from https://github.com/Free-Cities/Free-Cities/issues/170#issuecomment-290579968 (or #161)? 🤔

Bane70:
This is actually the only thing that I have changed since the new master copy came out, the accordion works fine as far as I can tell.

PR #193, opened 2017-04-01
WIP (no version number/release schedule)

freecitiesdev:
Next build no sooner than week of April 3 (realistically, midweek).

UI tweaks
Disaggregated the display of numerical stats from other display options, making it possible to enable and disable them independently.
Added links to slaves' individual menus to assign them to facilities directly.
Allowed reassignment of slaves between facilities without removing them from the first facility.
Arcology share purchases can now be made in lots of 10% if sufficient cash is available, limiting transaction costs.
Other minor UI improvements.

Arcology development changes
Reduced the pace of organic prosperity development for all arcologies; event impacts on prosperity are unchanged.
Added a powerful layer of cultural defensiveness to all arcologies under the complete control of their owners.
All cultural influence modifiers now affect the thresholds for organic influence in addition to the efficacy of targeted influence.

Slave stat changes
Refactored waists to be more granular, adding new attractiveness ranges and removing randomness.
Hormones will now affect waists; masculine hormones especially will broaden waists out to a fairly masculine level.
Corsetage pain, fear, and pregnancy effects are now all limited to extreme corsets, which also narrow waists much faster and farther.
Refactored faces to be more granular, removing randomness.
Advancing age will very slowly reduce facial attractiveness, though high health and trust can help protect against this aging.
Extremely high health and trust can actually improve facial attractiveness for slaves who are already beautiful.
Added several new potential effects for slaves under extreme mental stress, including slight losses of facial attractiveness.

Prostates
Slave's prostates are now statted; dicks and prostates generally appear together, but either can now be present without the other.
Several existing anal stimulation mechanics are now dependent on the presence of a prostate and can be suppressed by surgically removing it.
Conversely, implantation of a prostate will subject slaves without dicks to those anal stimulation mechanics that previously required one.
The organ farm is now capable of fabricating prostates for all slaves; the existing penis and testicle fabrications now come with a prostate.
NB: reactions to prostate implantation, effects on cum production pending; initial .prostate setting not complete.

Other updates
Added vector art for most collar options, by XXXX.
Updated many Racialist mechanics to properly support the Malay race, which was previously included in some but not all racial situations.
Enabled mixed race Supremacism and Subjugationism.
Added a policy to determine whether the PC is hosting social events, which is active by default.
New random event in which a young social climber flirts with the PC at one of the PC's social events.
New random slave event for trusting sadists when an Arcade is present.
Some very basic cleanup of old code for sanity and legibility.
[...]

PR #187, opened 2017-03-30
Switch TwineScript to JavaScript conditional operators

Folant:
The idea of replacement came after [this comment](https://github.com/Free-Cities/Free-Cities/issues/134#issuecomment-287714375) and the ongoing work of @Bane70. (e.g #174 etc)

Also fixed https://github.com/Free-Cities/Free-Cities/blob/cc40625e15d57a56a10987d5492bd40c043b6ab2/src/uncategorized/newSlaveIntro.tw#L654

I tried to cover 100% [patterns] of it, but work must continue at some time, of course

PR #194, opened 2017-04-01
Cherrypick of commits since #187

freecitiesdev:
Attempted partial resolution of a versioning issue I seem to have introduced with #193. Reapplies commits since #186-#185. Only known lost changes at this point are the javascript operators changed in #187 that directly conflict with subsequent changes. Apologies to Folant and anyone else inconvenienced; I'm about 90% sure I know where I went wrong and how to avoid doing so in the future.

If someone with more GitHub knowledge can easily see a more elegant resolution to this, jump in.


Bane70:
Ah I was wondering what happened. I seem to be not getting notifications of things on here, but they are showing up in my emails. I was unaware you had updated the master and I went and posted a fix that had a major conflict. looks like the facility arrays are totally gone besides maybe #185, from the little I've looked around and saLongTermEffects seems to be partly missing, errors while trying to build. Also the fix I did was something that should not of needed fixing as that was well tested a missing $i = _FLs for the stewardess, but I did have more than one edit of it up so maybe git got confused on the merge of that but shouldn't of have.

Folant:
`git reset --hard HEAD~2` restores last working state of the game ( to #191 ) locally for me, I will now try to figure out how to apply changes introduced after that.

Folant:
The most elegant solution I came to today, is to cherry pick #193 on my local branch, automatically replace all of the conditionals to js ---> it will leave only changes from @freecitiesdev WIP  ---> apply and test it then. Later we can think of how to do it on master, with 'revert' maybe.

Folant:
>Apologies to Folant and anyone else inconvenienced

That's my fault actually, I should have done this after release.

PR #197, opened 2017-04-02
Cybernetics, eugenics and PC skills.

Rodziel:
Includes:
Eugenics filter at recruiter, Cybernetics mod and PC skills.
Skills that are not included in your career choice can be trained from personal attention screen.

Something I changed is causing error in VaginaDescription widget for some test characters(Miss Anne, etc). Generated slaves don't seem to get this error. Haven't been able to track down the cause.


freecitiesdev:
This is outstanding and I'm now merging it. I will review it further in my next round of playtesting for lore compliance; shouldn't be anything serious.

Rodziel, you know these changes better than I do. Can you write up a changelog in the same style I usually use, for inclusion in the next update's changelog? I could do it if you'd like, but I'd probably miss changes.

PR #208, opened 2017-04-04
slaveAssignmentsReport.tw fix №2

Folant:


Bane70:
Ah thanks for correcting that. That was code from some testing I was going to be doing, but never got around to do it. I got to many copies of the game on my system, in one state or another. Time for me to delete 'em all, and work on one project at a time.

PR #228, opened 2017-04-06
Slave Template

Bane70:
Got rid of four of the five slave templates that where being used, and created a widget (BaseSlave). As the template in storyInit was missing a var or two (eyewear) and had some bits set to numbers instead of strings, also had to move it as a result to initNationalities as it wouldn't let me initialize $customSlave (startinit maybe don't like widgets? Maybe widgets and JS are not initialized as of this point.)
Generate XY Slave and $customSlave both set the vars values that are different to what they were. The rest were the same.


freecitiesdev:
> startinit maybe don't like widgets?

It's always something. Good work.

PR #234, opened 2017-04-07
Var fix BackwardsCompatibility.tw

Bane70:
Fix the lowercase .csec to be .cSec. also delled a couple of dupe entries.
Noticed that .induce is being added to the slaves as well but it is missing in the template., and it is also only being initialized in longSlaveDescription if it's not defined. Other than that its not being used.


pregAnon:
Figured I should start getting into the habit of using github.  .induce is safe to remove; it would be used in seBirth, but the content involving it is not implemented.  Odds are it slipped in when I moved the hair descriptions from longSlaveDescription to descriptionWidgets and did a little cleaning.

Bane70:
OK. Thanks.

freecitiesdev:
Pleased to see you here, pregAnon.

PR #242, opened 2017-04-07
fix pornfamespending?

goldifrocks:
I think both cases are meant to fire only with active spending, otherwise, a slave with prestige>=3 keeps getting the 'Further paid publicity cannot increase ...' message every week.

### DIFF
--- a/src/uncategorized/saLongTermEffects.tw
+++ b/src/uncategorized/saLongTermEffects.tw
@@ -3403,7 +3403,7 @@ and ($slaves[$i].clothes != "a slutty qipao") and ($slaves[$i].clothes != "restr
 
 <<set $pornFameBonus to 1>>
 <<if $studio == 1>>
-<<if $slaves[$i].pornFameSpending >= 0>>
+<<if $slaves[$i].pornFameSpending > 0>>
 	<<set $pornFameBonus += 0.2 + ($slaves[$i].pornFameSpending/10000)>>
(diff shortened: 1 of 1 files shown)


Bane70:
I'd say the first one definitely should be your way otherwise there's the dreaded divide by zero, which creates a NaN.

goldifrocks:
there is no NaN here, 0/10000 is 0 (according to my javascript console at least...).  this is mostly cosmetic, as the second condition includes the 'paid publicity cannot increase her fame', which for every slave with prestige>=3 repeats every week, even if they have no active pornfamespending :-(
that can be fixed by a different check, depending on whether a (very slow) increase in pornfame is wanted by @freecitiesdev even with pornfamespending=0.
this is just the variant I think to be the most consistent, i can make a different PR for just the message, if having the studio should be enough for slow pornfame increase and prestige chance :-)

Bane70:
Well that's good to hear about it not returning NaN, though infinity would return 0 but it would be infinite in them lol. 
Although arithmetically undefined, division by zero in JavaScript is valid and has well-defined results:
    1 / 0 yields Infinity
    (-1) / 0 yields -Infinity
    0 / 0 yields NaN
As a result, JavaScript will not throw an error when dividing by zero so it is up to the programmer to check whether the divisor of a fraction is not 0. Maybe Sugarcube is correcting it to be 0
(isFinite($slaves[$i].pornFameSpending/10000))

PR #244, opened 2017-04-08
Automated multiple organ and limb application.

DrNoOne0:
Adds the ability to have all completed organs and limbs automatically applied.

- Follows the same rules as manual application (i.e. costs & health losses).
- Will not discard organs or limbs that cannot be applied due to improper slave organs or limbs.
- Gives a single screen of text with all the slave reactions to their surgeries.  
- Does not give messages for ready but not applied organs.
- Discards organs and limbs for slaves you do not have.

Note: the current descriptions do not give the slave names, and can be confusing. surgeryDegradation.tw should be updated to lead with the slave's name.


freecitiesdev:
Outstanding.

PR #253, opened 2017-04-09
Compiler Changes (Please Read Before Accepting)

ObstacleCorpse:
I've received a few emails asking for a 32 bit Windows compiler, so I'm adding it (and others) to the repo. I have edited the Windows compile scripts to check the user's CPU architecture and choose the appropriate compiler. However, I have only tested this on one 64 bit computer running Windows 10.

I've uploaded a copy of my repo with the new compile scripts here:
https://mega.nz/#!aNAUWQjC!DRe1rPhgMvNza04wCN70a6Oir3SzA6NtBSpPrnsvtGU
I would encourage you to download the above and run the compile scripts. If they produce a working build of FC, I would assume everything is fine. If not, please let me know and do not accept this pull request.

Changelog:
Added TweeGo for x86 Windows.
Adjusted Windows compile scripts to choose the appropriate compiler for the user's CPU architecture.
Added TweeGo for x86 Linux.
Adjusted Linux compile script to use the new filename of the 64 bit Linux compiler.
Added TweeGo for x86 and x86_64 OSX.

Notes:
I don't have access to a Linux box at the moment, but I will edit the compile script to check the user's CPU architecture as soon as possible. For now, it continues to use the 64 bit version of TweeGo.

I don't have a Mac, and have no experience writing scripts for that platform. Though it shares the same underlying Unix structure, I don't want to create scripts blindly based on Linux and hope they work. The compilers are there if someone with a Mac wants to write the scripts.


HorselessHorseman:
How did they get ia32 processors? I thought nobody uses those.

PR #257, opened 2017-04-10
Fix rules assistant smart piercing gui

for-anonimity:
solves the issue with not finding span in the gui and also add missing fields to new rules.
changes to RAwidgets for taking "no default setting" into account


for-anonimity:
committed only one file, brb

PR #269, opened 2017-04-14
Restored lost content.

pregAnon:
Just restoring my lost descWidgets stuff that got reverted with the null descriptions.


freecitiesdev:
It's frustrating that you had to do that. I thought I'd corrected my workflow so that wouldn't happen again. I'll contact my Git experts to see where I'm going wrong.

PR #273, opened 2017-04-15
Change styles to use css classes, so that they can be customized

Stuffedgame:
This would be really useful to allow people to have a light colored css scheme.  If approved, I can follow this up with another commit to let people easily change to a light css scheme.

This was commit was created automatically by doing:

1. Search and replace "@@color:"  with "@@."
2. Search and replace "@@.rgb" with "@@color:rgb"
3. git diff | sed -ne 's/^[+].*@@[.]\([^;]*\);.*/.\1 { color: \1 }/pg' | sort |uniq

and put the output in mainStyleSheet.tw


freecitiesdev:
I'm cautiously supportive of this. The "cautiously" is due to there having been a serious issue with the implementation of the last massive find-and-replace scheme. Can you resolve the conflicts I see here and test this thoroughly? If that goes well, I'm inclined to accept this, since more options that don't affect people who don't use them are always acceptable.

johnflux:
I updated and tested.

johnflux:
I've updated again - it does fall out of sync quite quickly, so just let me know and I'll rebase or remake it.

freecitiesdev:
As long as this is testing fine, I'll merge it once it's conflicted out.

FCD

johnflux:
Okay I've just fixed the conflicts again and updated.

PR #330, opened 2017-04-26
Update SugarCube 2 to v2.18.0

ObstacleCorpse:
Release Notes:
https://www.motoslave.net/sugarcube/2/releases.php


devTools/tweeGo/targets/sugarcube-2/header.html
```diff
@@ -2,13 +2,13 @@
 <html data-init="no-js">
 <head>
 <meta charset="UTF-8" />
-<title>Free Cities</title>
+<title>SugarCube</title>
```
Forovakill: Is reverting this necessary? It's not really that important, but keeping it make it make the initial startup look cleaner. (Specifically, it causes startup to show the games title instead of just SugarCube, which looks more professional imo)

ObstacleCorpse: Ah, I see. No, reverting that isn't necessary at all.
I don't actually develop SugarCube 2; I just download the new versions and overwrite the old version if the release notes don't suggest incompatibilities. I didn't know people were making changes to the header, so I never diff the files before committing.

The latest push reverts the title back to Free Cities. Thanks for catching that.

PR #331, opened 2017-04-26
Initialization for upcoming universal rule, flat chests and a myriad …

pregAnon:
…of bug fixes.


Forovakill:
If i recall correctly, all other flat breast checks are less than 300, not less than or equal to.

PR #332, opened 2017-04-27
Update initNationalities.tw

Atacama74:


Atacama74:
Noticed "Swedish" was missing from the Europe case, so I added it.

PR #345, opened 2017-04-29
my first contribution - a collection of fixes and usability improvements

fcpmanon:
thanks for all your efforts!  these are some of my personal edits that I hope will be useful to others

please let me know if you have any questions about the changes - I've described most of them below:

brothelAssignmentScene.tw - cleanup, even though the passage is currently unused
mainStyleSheet.tw - change sizing of main passage to use up to 90% of the available space - a usability improvement for users who zoom out / reduce font size
researchLab.tw - fix a typo, bugs in hiring logic, and add stockpile counts to main display so users don't have to go into purchase screen to check current stockpile
bodyModRulesAssistantSettings.tw - even if global $seeDicks toggle is disabled, allow users to modify dick piercing rules if they were previously set (NG+ carry-over, for example)
clinic.tw - extend facility description text when nurse is not present to inform users that blood treatment is not active
clinicReport.tw - changed automatic release threshold to wait until genome damage is completely removed, instead of kicking them out when it is below 20
hgApplication.tw - use Math.ceil function instead trunc, rounding up to nearest integer, to avoid possibility of zero skill increases when effectiveness is low
longSlaveDescription.tw - use Math.ceil function instead of trunc, rounding up to nearest integer, when reporting current genome damage level
manageArcology.tw - when buying arcology ownership with reputation, restore demand rating so share values remain consistent with cash purchase method; refer to "engineering training" instead of "engineering career" because the skill may be from training rather than starting career
randomIndividualEvent.tw - check for non-permissive release rules to conditions for aphrodisiac addict event as the event text mentions not being allowed to masturbate
saRelationships.tw - replace "is" with "==" comparison operators for consistency
spaReport.tw - consolidate duplicate career bonus checks for Attendant; adjust mindbreak cure text to not mention "weeks" as it could happen in a single week; move career bonus outside of slave trust check as bonus should also apply to trusting slaves; better flaw softening text - indicate type of flaw corrected (sexual/behavioral) and use same spacing as mindbreak cure; don't kick slaves out of the spa while they still have unresolved flaws
raWidgets.tw - replace "None" condition for facilities and assignments with "All"; fix various bugs in auto rules assistant application so that selecting include & exclude conditions of more than one type works better and so that auto-assignment to facilities correctly matches their qualifications; suppress smart piercing adjustments if the previous setting was not actually changed (as a result, piercings will cycle through the modes [libido, +XX, +YY] each time the rules are applied, if the slave has not achieved the target yet); consolidate fertility settings changes; fix text for accessories changes
slaveAssignmentsReport.tw - adjust line breaks and fix auto rules assistant applying to facilities (and suppress the auto-RA output for slaves whose assignments are not visible)
slaveSummary.tw - add new "Markings" abbreviation to slaves that have markings; remove the "Dr:None" case so that searching for "Dr:" allows users to find which slaves have active drug settings; add new "RA-Exempt" indicator for slaves that have been exempted from the rules assistant; display genome damage in clinic facility for slaves in the facility as well and use Math.ceil (round up) instead of trunc (round down); remove typo '' before nationality section; add "weeks" text to sentence display; correct qualifications for quick assignment shortcuts from penthouse summary screen
slaveInteract.tw & assignWidgets.tw & remoteSurgery.tw & surgeryDegradation.tw & various other files - use assignJob and removeJob widgets in places where slaves' assignment property was being set directly to avoid situations where slaves are removed from facilities without removing their IDs from the corresponding facility array (the remote surgery code was responsible for some of these glitches, particularly for the blind, amputate, and install prosthetic interfaces surgeries); correct qualifications for assignment shortcuts; move fucktoy preferences into a dynamic section that appears (and disappears) as needed (via new widget); make fucktoy preferences adjustable for slaves in master suite, since they still affect them there; clean up vertical bar dividers in cases where some of the mini-scene shortcuts are unavailable; adjust spacing and line breaks; save changes to active slave when exiting interaction screen via facility assignment shortcuts (usability improvement); dynamically update sex act mini-scene shortcuts when removing chastity belts (usability improvement); also accounted for some edge cases with agents and agent companions
brothel.tw - added new option in the facility screen to deactivate/re-activate the aphrodisiac injection systems after they have been installed; when deactivated, they still incur maintenance costs, but only 10% of the cost of keeping them active; since the injector systems prevent slaves from gaining sex skills due to giving them extreme aphrodisiacs rather than normal dosage, this is a very useful toggle


fcpmanon:
oh, I should also mention that the removeJob widget now sets sentence = 0, and that the assignJob widget can now be used for non-facilities assignments also, as it will set assignmentVisible appropriately for the facilities, and it still calls removeJob first so it will always take care of updating the facility-specific arrays

fcpmanon:
Few more small typo bugs found - will push another patch to this branch soon.

PR #349, opened 2017-04-29
code refactoring

Frodfurgurn1991:
Should all be inconsequential changes.

### DIFF
--- a/devTools/tweeGo/targets/sugarcube-2/userlib.js
+++ b/devTools/tweeGo/targets/sugarcube-2/userlib.js
@@ -513,6 +513,26 @@ window.hasSurgeryRule = function(slave, rules) {
 	}
 };
 
+window.lastRuleFor = function(slave, rules, what) {
+	if (!slave  || !rules || !slave.currentRules)
+		return null;
+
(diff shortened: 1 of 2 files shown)


Frodfurgurn1991:
ruleSlaveSelected and Excluded explicitly check if rule.selectedSlaves is "truthy".  There are several places in raWidgets.tw that also explicitly check if ndef rule.selectedSlaves

I'd really like to eliminate all of those cases...

fcpmanon:
Can you look over my RA changes?  See separate pending merge request.

Frodfurgurn1991:
@fcpmanon What exactly do you want me to check?  Your PR doesn't seem to conflict with mine in anyway.

In any case, I haven't dabbled too much into the RA code just yet... I plan to do a biiig change sometime in the future, if I find the time for it.

fcpmanon:
Your previous comment mentioned changes to the RA widgets.  I made some changes there, so I was hoping you could review since you have ideas about improvements there.

Frodfurgurn1991:
I also have a question pertaining cumAmount... Why is it that slaves with more than 50 or less than -50 health produce the same amount?  While the rest produce less?

PR #351, opened 2017-04-29
Fix dupe slaves after surgeries in facilities

Bane70:
Also fixes a bit of text that was just being displayed for Fuckdolls instead of all surgeries.

#345 The conflicting surgeryDegradation.tw posted doesn't fix the facility issue as the removeJob is only working for fuckdolls, and I'm sure that agents should not be a fuckdoll, though I didn't code the removal of agents in.


fcpmanon:
Aha.  Good catch.  This probably explains some of the ID switching I noticed too.  I bet I have a bunch of duplicates in my list due to surgeries.

fcpmanon:
Never mind my last comment.  This only affects the facilities lists, not the main slaves list.

Bane70:
This was an error that I thought I had fixed a while back but either it got reverted back, a month or so ago or I had forgotten to post it, due to things getting reverted. It was not what I was expecting to see when I was looking into a bug report lol.

fcpmanon:
I've incorporated the changes you made here into my pending merge request, with a few other additions (added a separate defensive check for assignments not eligible for surgery; typo fixes, deduct prosthetics interfaces from stockpile when attaching, replace some 'to' and 'is' with = and ==)

Bane70:
OK, then I'll close mine then. Thanks.

PR #352, opened 2017-05-01
Free Range Dairy Assignment scene tweaks

prndev:
This commits contain tweaks and bugfixes regarding the Free Range Dairy Assignment scene as well as a bugfix regarding an anal fetishist's reaction to prostate implantation.


prndev:
@fcpmanon Good points. I want to regard them in my next pull request.

PR #357, opened 2017-05-02
Rules Assistant usability & assignment code improvements

fcpmanon:
more fixes for (re-)assignment code, including moving on-assighnment facilities rule changes into the respective facilities' weekly code; more Rules Assistant fixes and improvements, including enabling the until-now-disabled feature of the RA that allows it to re-assign slaves that are in facilities, a condensed and (hopefully) more understandable activation conditions section, and mutual exclusion of incompatible conditions; added the ability to change the priority level of existing rules; various bug fixes, including defensive code to deal with nulls in the $defaultRules array; and last but not least, adding or importing rules re-numbers all rules' internal IDs to avoid some glitches that occur when multiple rules end up with the same ID; and some incidental cleanup in userlib.js to support all of this

### DIFF
--- a/devTools/tweeGo/targets/sugarcube-2/userlib.js
+++ b/devTools/tweeGo/targets/sugarcube-2/userlib.js
@@ -129,31 +129,59 @@ if (typeof SlaveStatsChecker == "undefined") {
 	window.SlaveStatsChecker = SlaveStatsChecker;
 };
 
+window.removeFromArray = function(arr, val) {
+	for (var i = 0; i < arr.length; i++) {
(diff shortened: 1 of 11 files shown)


corncobman:
What happens to rules that are applying to slaves when the ID gets changed, do the IDs change accordingly?

fcpmanon:
Good catch.  I added code to re-sync slaves' $currentRules IDs after renumbering the IDs in $defaultRules.

Also it seems I uploaded the wrong version of ruleAssistant.tw before.  Looks good now.

PR #358, opened 2017-05-02
BackwardsCompatibility - oversight correction

corncobman:
Unbork version number checking


fcpmanon:
looks like no conflict with #360

PR #359, opened 2017-05-02
EditSlaveCheatDataCleanup - Add missing birthWeek

corncobman:


freecitiesdev:
Thanks. If you don't mind, keep an eye out for .birthweek becoming a string. I think there's still a bug somewhere that occasionally produces that outcome. Haven't been able to find it yet.

PR #360, opened 2017-05-02
0.10.1 backwards compatibility update

fcpmanon:
after loading an old save, go to the options menu and click the new link at the top


fcpmanon:
It would be nice to automate silently loading the Backwards Compatibility page after loading a save, but I don't know of a way to hook that up.  I don't see anything in Backwards Compatibility that would be harmful to run on an up-to-date game.

PR #376, opened 2017-05-04
Hoods, clits & misc.

Rodziel:
Enabled foreskins for pure females. 
Added descriptions for foreskins. 
Added descriptions for extra 2 clit sizes.
Enabled dick growth drugs for pure females. Will affect clit size.
Changed few links in arcology management so they link back to management screen instead of main. It was annoying.
Made separate activation for sidebar cheats in options since they clutter the menu.

### DIFF
--- a/src/npc/fFeelings.tw
+++ b/src/npc/fFeelings.tw
@@ -490,20 +490,36 @@ My favorite part of my body i<<s>>
 
 <<switch $activeSlave.drugs>>
 <<case "penis enhancement">>
-	<<if ($activeSlave.balls == 0)>>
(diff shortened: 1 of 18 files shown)


freecitiesdev:
Can I get the arcology links commit dropped out of this? I'd like to accept the other 3 commits but I can't seem to figure out how to do that.

I understand the frustration, but what I'm trying to move towards here is a system where the arcology interface is being used from the main screen. Since the terrible Sugarcube workaround for setter links created from arrays (like $sectors or $slaves) can't handle pass(), it has to be one or the other for now. These changes will make it just as frustrating for someone trying to use the arcology interface from Main.

Rodziel:
Reverted the changes, though I don't see the problem.
I only changed links for buying/selling sectors, facilities still linked back to main.

freecitiesdev:
New UI toggle is temporary.

PR #381, opened 2017-05-05
Use cat instead of strings, for better windows compatibility

Stuffedgame:


Stuffedgame:
The check:  " Check for duplicate words "

Finds around 30-40 hits.  I didn't have time to make a patch to fix these, so it would be good if someone else could.

Example:

[Duplicate words] src/utility/descriptionWidgets.tw:7107:        $activeSlave.slaveName's jacket barely closes over $possessive pregnant belly leaving leaving its' buttons threatening to pop.

> leaving leaving

PR #398, opened 2017-05-05
Add an expression parser for the RA

Frodfurgurn1991:
Every wanted to use a condition for the Rules Assistant that you couldn't represent?  Perhaps you wanted to use an "or" instead of an "and" or perhaps you wanted to check more than one variables.

Well, this PR is for you.  It adds one more "activation" type, called "custom", that can be any expression!  The language used is based on C and JavaScript, so "&&" is and, "||" is or and so on.  Do note that equality is just "=" since no  assignment can be done.  This may change in the future.

However, this PR also goes a step beyond just that.  It tries to integrate, as much as possible the current activation types and UI with this new code.  As such, when you set a condition through the existing interface, this will actually also be made available through the "custom" tab and vice versa, if the expression given is something that the UI can represent.


Frodfurgurn1991:
Oh, this is backwards incompatible.  Also, keep in mind that it breaks "Import Rule" and "Export Rule".  Does anyone actually use them?

freecitiesdev:
> Oh, this is backwards incompatible.

Would it be too onerous to add backwards compatibility? Based on my very quick review, it looks like it should be possible.

Frodfurgurn1991:
It definitely is, wait a bit.

Frodfurgurn1991:
Ok, this should work.

freecitiesdev:
RA menu is throwing **Error: <<if>>: bad conditional expression in <<if>> clause: Cannot read property 'id' of undefined.**

Mouseover refers to **<<if isSimpleCondition($currentrule...**

Frodfurgurn1991:
Welp... Do you have any idea why?  I mean, the code in BackwardsCompatibility.tw should be working...

Frodfurgurn1991:
Ok, so I see what is going on... BackwardsCompatibility.tw is something that is applied "on request" at the options screen...  Is there something more that I should do to make the change "automatic" ?

freecitiesdev:
BackwardsCompatibility.tw fires automatically when NGP is applied.

freecitiesdev:
And all exclusions seem to be broken. All rules are applying to all facilities regardless of rule settings.

Frodfurgurn1991:
Ok, I think I found out why.  Just to check though, it should not just be applying the rules, it should also be fucking up the output, like

> slave is now wearing $args[0].clothes.
> slave has been given a $args[0].collar collar.
> slave's living standard has been set to _combinedRule.livingRules.

and so on.

freecitiesdev:
It's doing both for slaves in facilities when facility exclusions should be applicable.

Frodfurgurn1991:
Yes.  The problem is that, when no rules are applicable for a slave, it still tries to apply the empty _combinedRule object, and that horribly breaks things.  For a quick check, try adding an `<<if (def _combinedRule.clothes)>>` around the code that applies rules, i.e just after the `<</for>>` and just before the end of the `<<DefaultRules>>` widget and see if that fixes stuff.

Frodfurgurn1991:
basically do https://github.com/Frodfurgurn1991/Free-Cities/commit/62e5c22a5361fc560bfcf4c3b463fe821e49c804 and see if that fixes it.

PR #420, opened 2017-05-08
Change the way RA rules are applied

Frodfurgurn1991:
So, right now, for each slave, the RA loops through all the rules that apply to her and makes the changes specified.  This has several consequences.  First of all, if more than one rules act on some "attribute", like say the clothes, you'll get messages like "x is now wearing this. x is now wearing that".  Secondly, it's kinda inefficient.  This doesn't matter that much because you generally have a small number of rules, but still.

What is PR does instead is that it merges all the applying rules into one, and then acts according to that.  So no more silly doubled messages and no more "last...Rule" functions!

Not everything is merged though.  In particular, jobs and facility assignments are still done the "old way", inside a loop.  So you may still see doubled messages for that.

In general though, this results both in simpler code and better messages.

This is not the only change this PR makes though.  It also eliminates rule variables that were never used and combines some other ones.


Frodfurgurn1991:
This will conflict with #398, since they both change initRules.tw etc.  That PR is meant to be pulled first.  After that, I'll fix the conflict if you want.

fcpmanon:
The .choosesOwnClothes variable tracks whether the slave should change her clothes again in later weeks, since once she chooses an outfit for the current week her clothes variable will be set to a specific outfit.

Frodfurgurn1991:
That may be the case for the slave itself.  However, when talking about rules, the .clothes variable of a rule will never change through the week.

fcpmanon:
Ah, I see.  You still set the slave's own chooses variable in raWidgets.  Thanks for clarifying.

freecitiesdev:
> That PR is meant to be pulled first. After that, I'll fix the conflict if you want.

Fire at will.

Frodfurgurn1991:
Looks ok now.

Frodfurgurn1991:
fuck, I messed that up

freecitiesdev:
Reposting from #398 because this is the only other PR which has changed the RA. Pretty sure it's that one, but in any case:

freecitiesdev:
Reverting this PR for now. At minimum, it breaks exclusions, applying rules to slaves in facilities regardless of facility exclusion settings, producing all kinds of issues. See #429.

Frodfurgurn1991:
Moving the #398 discussion here since this is the relevant PR.

So, any luck with that fix?

Frodfurgurn1991:
@freecitiesdev ping

PR #422, opened 2017-05-09
Use the <<Master>> macro where appropriate

Stuffedgame:
Try to create some consistency here, and fix a whole load of bugs.

Note: Please use the sanityCheck and fix the bugs that it catches.


freecitiesdev:
Does the sanity checker have a readme somewhere?

Stuffedgame:
I added a bit of readme here: https://github.com/Free-Cities/Free-Cities/pull/423

Anything in particular you want to know?

PR #423, opened 2017-05-09
More fixes found by sanity checker

Stuffedgame:


freecitiesdev:
Still not getting sanityCheck. I think you may need to dumb your explanation down a level. On Windows, it appears as an extensionless file. Is the idea to run it as a batch file like compile.bat?

Stuffedgame:
Ah sorry.  If you right click in the folder, you should get a 'git bash' option, like:  http://guides.beanstalkapp.com/version-control/git-on-windows/git-bash.png

Chose that 'git bash'   and then type in:

bash sanityCheck

Does that work for you?

freecitiesdev:
No such option appears.

Stuffedgame:
It's usually installed when you install git.  You could try installing git for windows:

https://git-for-windows.github.io/

I think by default it gives you those menus.

freecitiesdev:
That's done it. Thanks. I've been using GitKraken exclusively, FWIW, and didn't have git for windows installed prior. Worth adding to the readme.

PR #424, opened 2017-05-09
Economic Tabs

megas4o:
Toggleable tabs in the Economics Report

### DIFF
--- a/src/events/intro/introSummary.tw
+++ b/src/events/intro/introSummary.tw
@@ -171,6 +171,15 @@ of the slave girls will have dicks.
 
 <br>
 
+Economic Tabs on weekly reports are
+<<if $useTabs == 0>>
+@@.red;DISABLED@@. [[Enable|Intro Summary][$useTabs = 1]]
+<<else>>
+@@.cyan;ENABLED@@. [[Disable|Intro Summary][$useTabs = 0]]
+<</if>>
+
+<br>
+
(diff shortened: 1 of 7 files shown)


freecitiesdev:
I really like this. I feel like it needs a little more usability, though, since it can add clicks as it is. Two feasibility questions:

1. Would it be possible to remember which tab we're on, week to week? That way players who care about one tab wouldn't have to select it over and over.

2, Can we bind Q and E to cycle tabs like they currently cycle slaves' individual menus and rulesets?

megas4o:
It can be reworked to add both keyboard and tab functionality. Both of my projects were done in a day, since it's all I had free. Will look into it the next time I have time.
 

    On Wednesday, May 10, 2017 3:24 AM, freecitiesdev <notifications@github.com> wrote:
 

 I really like this. I feel like it needs a little more usability, though, since it can add clicks as it is. Two feasibility questions:   
   - Would it be possible to remember which tab we're on, week to week? That way players who care about one tab wouldn't have to select it over and over.
2, Can we bind Q and E to cycle tabs like they currently cycle slaves' individual menus and rulesets?—

freecitiesdev:
It should be straightforward, I agree. I just don't know enough HTML to point Mousetrap to the right <div> without going hunting for documentation, etc., so it'll be helpful if you can do that. Mousetrap lives in userlib.js, of course.

Issue #397, opened 2017-05-05
Build number is wrong


MegaPramuk:
Yesterday, I've built from master and saw it has 1005 commits already (according to the build number portrayed in the main menu at least) but after pulling and building again today, it's still 1005. 
I don't really understand the structure of this project (or rather, what exactly .tw files are) but if I'm not mistaken, the variable for this is declared in ./src/init/startInit.tw.
There should be some automation going on for the builds instead of hardcoded values.
With `git rev-list HEAD --count` one can easily find out the current commit count. And with sed (in ./compile at least) one could change $releaseID.

freecitiesdev:
$releaseID is being looked at.

PR #427, opened 2017-05-09
Build number replacing

MegaPramuk:
This PR addresses #397.
No batch file modifications because I don't have any Windows machine, so I wouldn't be able to test it.

### DIFF
--- a/compile
+++ b/compile
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+# Find and insert current commit
+COMMIT=$(git rev-list HEAD --count)
+sed -i "s/COMMIT/$COMMIT/" ./src/init/storyInit.tw
+
+
 # Will add all *.tw files to StoryIncludes.
 ./sanityCheck
 rm -f src/config/start.tw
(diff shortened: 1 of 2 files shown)


freecitiesdev:
Outstanding.

Frodfurgurn1991:
It looks like @freecitiesdev was using that number in a number of places though, to check if the build was recent enough.  Using a commit sha does not allow you to do that unfortunately...

MegaPramuk:
It's not the commit sha1 that's being used.
It's the count of commits.
Only checks that are with `$releaseID` are whether it's bigger or smaller than 1000. So, no checks have been affected.

Frodfurgurn1991:
Sorry, you are right.  That's a great change then!

freecitiesdev:
This doesn't seem to be working at the moment.

MegaPramuk:
Weird, I just tested it out and it works on my machine.

What's the build number for you? Does it say COMMIT?

freecitiesdev:
Well, you can see how I have it now:
`<<set $ver = "0.10.2", $releaseID = COMMIT>>
<<if ndef $releaseID>><<set $releaseID = 1007>><</if>>'
Compiling from a straight download of the repository and firing compile.bat produces FC.html; opening this shows $releaseID as 1007. This is on a Windows box.

PR #430, opened 2017-05-09
Quartet of new makeup flavors.

pregAnon:
Also comes with restored old collection facility content reworked for the dairy and initialization for custom titles.


freecitiesdev:
> restored old collection facility content reworked for the dairy

What, exactly? Can't see it in the files changed.

PR #441, opened 2017-05-11
fix the RA, agents, slave interaction and more (update: now includes #442 above)

fcpmanon:
important changes highlighted:

**fix numerous bugs with RA and special slave exclusions and facility assignments (among other issues, several uses of _currentRule needed to be changed to _combinedRule)**
remove RA "auto-assign to get milked" for now as haven't implemented lactation checks for it (auto-assign to Dairy facility is still available and does have appropriate checks implemented)

fix long slave description when slave's clothes are set to "choosing her own clothes"

**fix personal attention to stay on HG after switching with HG Suite slave**
display text about changing personal attention when assigning current PA slave to a non-penthouse job

**set agents & agent companions to be RA-Exempt when put on assignment**
**hide agents & agent companions when displaying slave summary lists to prevent players from accessing their slave interact pages while they are on assignment**
allow slave lovers/wives of agents to be sent to live with the agent from any assignment, instead of only when they are in the penthouse
sanity check during weekly neighbors development report for agents that aren't properly on assignment (backwards compatibility - should not be possible to get into this state without an old save)

add <<z>> lisp widget to <<Enunciate>> widget, and use it to streamline agentCompany.tw

return to Manage Arcology from Neighbor Interact instead of falling out to Main

fix how slaves set to choose own assignment are displayed on summary list and **make them keep choosing each week as long as not explicitly assigned another job by the player or RA**

lots of cleanup of slave interact screen for consistency
**change "Transfer to" facilities links on slave interact screen to return player back to origin facility/main instead of following the slave to the destination facility**

**allow players to view and modify slaves' rules while they are in facilities**

**added assignment of a sector for the Pit, if missing, to the backwards compatibility update**
fix typo in empty Pit facility description

clean up mercs naming event
fix spacing in saLiveWithHG


Frodfurgurn1991:
> restore RA-anon's parser-based RA implementation

I don't know what you mean by that, the parser was never reverted.  Also, more importantly, could you please not restore _my_ changes?  And let me do it myself?

fcpmanon:
FCDev reverted rulesAssistant and raWidgets in #429 so it did not call the custom condition parser; I also accidentally removed the associated backwards compatibility lines in #432 and restored them in #433

For this PR, I undid FCDev's reverts of rulesAssistant and raWidgets and then made a large number of edits as I encountered problems during playtesting over the course of the day ... and attempted to address several of the open issues

I can fully back out the changes for rulesAssistant and raWidgets if you want, but perhaps you could review them first?

Frodfurgurn1991:
> FCDev reverted rulesAssistant and raWidgets in #429 so it did not call the custom condition parser;

But he didn't.  The only places the custom rules mesh with the other logic is [here](https://github.com/Free-Cities/Free-Cities/blob/master/src/utility/raWidgets.tw#L2417-L2428), [here](https://github.com/Free-Cities/Free-Cities/blob/master/src/utility/raWidgets.tw#L4721), [here](https://github.com/Free-Cities/Free-Cities/blob/master/src/uncategorized/rulesAssistant.tw#L150) and [here](https://github.com/Free-Cities/Free-Cities/blob/master/src/uncategorized/rulesAssistant.tw#L213-L238) and all these lines are very much in the game as you can see.

What FCDev did revert is the changes are you are trying to restore, which is #420 (blaze it).  And I'd be very happy if you let me restore them myself...

fcpmanon:
OK, I'll close this for now and reopen after merging on top of your changes

fcpmanon:
reopening -- I've merged the changes from #442  (still pending) with all my earlier work (as described above) and added some additional variables to exportRule.tw

Frodfurgurn1991:
You know what would be awesome?  If you could break up that huge commits that does all the things into many different ones that only do one thing.  And also, squash all the other tiny commits into some larger ones wherever that makes sense...

fcpmanon:
that's a good idea, but unfortunately I have no idea how to do that other than by manually redoing everything piece by piece?  if there are specific parts that you or FCDev want me to remove, I can do that

Issue #410, opened 2017-05-06
You cannot assign girl to HG suite from her own page, but you can from RA


lowercase-donkey:
| RA: Apply to facilities: | All | The Head Girl Suite | The Brothel | The Club | The Arcade | The Dairy | The Servants' Quarters | The Master Suite | The Schoolroom | The Spa | The Clinic | The Cellblock 
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- 
Unique: Transfer to: | | | Brothel | Club | Arcade | Dairy | Servants' Quarters | Master Suite | Schoolroom | Spa | Clinic | Cellblock | 

Oversight, or is there a reason for this?

fcpmanon:
Since only one slave can be assigned to the HG Suite, it's of limited value to have a shortcut to it.

lowercase-donkey:
Oh, I agree.  I just think the shortcut should be removed from RA as well.  It makes even less sense to have it there.

It makes sense to have a rule that can apply to HG suite, but not to transfer girls there.

Remove or add, but keep it the same.

fcpmanon:
I wouldn't remove it from RA.  I could imagine setting up a pair of rules in the RA to rotate slaves into the HG Suite and out of the HG Suite to some other assignment each week or after some other conditions.

PR #442, opened 2017-05-11
Various RA changes and fixes

Frodfurgurn1991:
More details in the commits, but in short, tries to fix some edge cases of #398 and redoes some of the work of #420 but not the merging of rules.

PR #443, opened 2017-05-11
fix the RA, agents, slave interaction and more -- now ready for merging

fcpmanon:
important changes highlighted:

**fix numerous bugs with RA and special slave exclusions and facility assignments (among other issues, several uses of _currentRule needed to be changed to _combinedRule)**
remove RA "auto-assign to get milked" for now as haven't implemented lactation checks for it (auto-assign to Dairy facility is still available and does have appropriate checks implemented)

fix long slave description when slave's clothes are set to "choosing her own clothes"

**fix personal attention to stay on HG after switching with HG Suite slave**
display text about changing personal attention when assigning current PA slave to a non-penthouse job

**set agents & agent companions to be RA-Exempt when put on assignment**
**hide agents & agent companions when displaying slave summary lists to prevent players from accessing their slave interact pages while they are on assignment**
allow slave lovers/wives of agents to be sent to live with the agent from any assignment, instead of only when they are in the penthouse
sanity check during weekly neighbors development report for agents that aren't properly on assignment (backwards compatibility - should not be possible to get into this state without an old save)

add <<z>> lisp widget to <<Enunciate>> widget, and use it to streamline agentCompany.tw

return to Manage Arcology from Neighbor Interact instead of falling out to Main

fix how slaves set to choose own assignment are displayed on summary list and **make them keep choosing each week as long as not explicitly assigned another job by the player or RA**

lots of cleanup of slave interact screen for consistency
**change "Transfer to" facilities links on slave interact screen to return player back to origin facility/main instead of following the slave to the destination facility**

**allow players to view and modify slaves' rules while they are in facilities**

**added assignment of a sector for the Pit, if missing, to the backwards compatibility update**
fix typo in empty Pit facility description

clean up mercs naming event
fix spacing in saLiveWithHG


src/utility/raWidgets.tw
```diff
@@ -3191,49 +3176,204 @@ Your brand design is ''$brandDesign.''
...
-<<set _currentRule = $defaultRules[_r]>>
-<<if !ruleApplied($args[0], _currentRule.ID)>><<continue>><</if>>
+	/% skip inapplicable rules %/
+	<<if !ruleApplied($args[0], $defaultRules[_r].ID)>>
+		<<continue>>
+	<</if>>
 
-<<if $args[0].fuckdoll == 0>>
+	<<set _combinedRule = mergeRules([_combinedRule, $defaultRules[_r]])>>
+
```
Frodfurgurn1991: Are you not checking for special slaves?

fcpmanon: I moved the special slave exclusion code into CheckAutoRulesActivate since that gets called before DefaultRules everywhere except when you "apply RA" from the slave interact page

fcpmanon: I remembered incorrectly.  CheckAutoRulesActivate _does_ get applied from slave interact.  DefaultRules is called without CARA only in newSlaveIntro and "apply RA" link from main

in newSlaveIntro it probably _should_ be called .. and the "apply RA" link from main only affects slaves in the penthouse and I'm not sure if the intent there is to just reset them back to whatever rules were applied before or to actually do an update-check (in which case CARA should be added)

Frodfurgurn1991: Yes, but that excludes the whole of the rule.  Previously the rule was applying normally, except that for special slaves it didn't change their assignment.

src/utility/raWidgets.tw
```diff
@@ -3191,49 +3176,204 @@ Your brand design is ''$brandDesign.''
...
-	<br>$args[0].slaveName is now allowed to choose her own clothes.
-	<<set $args[0].choosesOwnClothes = 1>>
+	<<case "mastersuite">>
+		<<if ($masterSuite > $masterSuiteSlaves) && ($args[0].devotion > 20) || (($args[0].devotion >= -50) && ($args[0].trust < -20)) || ($args[0].trust < -50)>>
+			<<if ($args[0].assignment != "serve in the master suite")>>
+				<br>$args[0].slaveName has been automatically assigned to $masterSuiteName.
+				<<assignJob $args[0] _combinedRule.assignFacility>>
+			<</if>>
+		<<else>>
+			<<if ($args[0].assignment == "serve in the master suite")>>
```
Frodfurgurn1991: I think that all these assignment checks before `<<include "Rules Facility Remove">>` are not needed, since "Rules Facility Remove" already checks the assignment of `$activeSlave` before removal.

src/utility/raWidgets.tw
fcpmanon: my interpretation of that option was that it should apply to the whole rule, like the specific slaves include/exclude list and everything above does .. I could see where someone might want an exclusion rule that only affects the assignment section but in that case it should be a sub-option that only appears when you actually select a setAssignment or assignFacility

src/utility/raWidgets.tw
fcpmanon: quite possible - those RFR calls were there before this PR and I didn't analyze them in detail

fcpmanon:
should I add CARA before the DefaultRules in newSlaveIntro and in Main?

Frodfurgurn1991:
What is CARA?  There's no such file anywhere...

src/utility/raWidgets.tw
fcpmanon: also, are you sure special slaves exclusion worked that way previously?

the way I read it, the old code if _exclude was true did a <<continue>>, thereby skipping the whole rule without applying any of its effects (assign changes happen first in DefaultRules) .. however, the _exclude special slaves check with the massive IF statement was broken (it was somehow true for a slave that wasn't in any of those roles), but I rewrote it differently and it now seems to work

fcpmanon:
I mean CheckAutoRulesActivate

fcpmanon:
since slaves are generated with .currentRules = [], I think the <<DefaultRules>> in newSlaveIntro does nothing at the moment

I'm not sure of the intent for the "apply RA" link in penthouse

Frodfurgurn1991:
In any case, I have this https://github.com/Frodfurgurn1991/Free-Cities/tree/ra-merge-again branch.  Now, realistically, I'd open a PR.  Seeing how you are basically including most of the changes in yours... I don't know what to do...

fcpmanon:
If you merge that branch into your #442 branch, I can then rebase my changes here on top of them, right?

fcpmanon:
I'm now convinced that at least the missing CheckAutoRulesActivate in newSlaveIntro is a bug/oversight, because if you have universal rules set to apply RA to new slaves, _that_ code (in AddSlave widget in miscWidgets.tw) _does_ call CheckAutoRulesActivate before DefaultRules

Frodfurgurn1991:
Well, it would probably conflict with everything you did, so...  you can still rebase though on it though, you just have to add my fork as a remote and do the rebase off of the branch there.

fcpmanon:
I'm working on it.

fcpmanon:
Not sure about your assertion here:

/% In case a slave has no rules applying to them, _combinedRule would (should)
be an empty object, and that would wreck havoc to the rest of the code.  So, we
check if a rule attribute, any one, is defined and otherwise skip everything.
%/

We want to know if a rule has non-default non-assignment settings so we can skip a lot of individual checks.  You check .clothes (randomly) to determine this.  But what if _combinedRule has non-defaults for everything _except_ .clothes?

I don't see an obvious way to determine if ALL non-assignment non-condition variables are default with a single check.  So, I'm going to skip that for now and go with the safer (and slower) def checks for every rule attr.

Frodfurgurn1991:
`_combinedRule` will be either an empty object or a "full" one, full meaning that it will have all the variables that a rule will have.  And that is all that we care for.  If it has one variable, it is gonna have all the other too, it doesn't matter if they are default or not, we just want them to not be undefined.

fcpmanon:
OK, I am not sure if that worked exactly right... I rebased on your other branch, and then force-pushed the result to this branch

it seems to have now forgotten that it was based on the ra-fix branch before; but I brought along the changes from there to this branch anyway

the only files that have changed as a result of this rebasing are:

exportRule.tw, initRules.tw, rulesAssistant.tw and raWidgets.tw

Attached is a diff showing the changes in those files to make it easier for you to review and check if I made any mistakes in the merging process.

[patch.diff.txt](https://github.com/Free-Cities/Free-Cities/files/994015/patch.diff.txt)

fcpmanon:
I don't quite understand how the all-or-nothing logic works, but I will add that extra if check back in now

src/utility/raWidgets.tw
```diff
@@ -1539,19 +1541,20 @@
 	<<link $HGSuiteNameCaps>>
 		<<set $currentRule.assignFacility = "hgsuite">>
 		<<set $currentRule.setAssignment = "none">>
-		<<set $currentRule.facility.delete("hgsuite")>>
-		<<if $currentRule.facility.length == 0>>
-			<<set $currentRule.excludeFacility.push("hgsuite")>>
+		<<if $currentRule.excludeFacility.length > 0>>
```
Frodfurgurn1991: why was this added? same with all the other `length` checks

src/utility/raWidgets.tw
```diff
@@ -3281,210 +3294,160 @@ Your brand design is ''$brandDesign.''
 <<widget "DefaultRules">>
 <<set _combinedRule = {}>>
 <<for _r = 0; _r < $defaultRules.length; _r++>>
-    <<set _currentRule = $defaultRules[_r]>>
```
Frodfurgurn1991: I'd rather you keep this, I don't wanna have to track down any more code that relies on global or local variables like that.

src/utility/raWidgets.tw
```diff
@@ -3281,210 +3294,160 @@ Your brand design is ''$brandDesign.''
...
-        <<if ($args[0].choosesOwnAssignment == 0)>>
-            <<assignJob $args[0] _currentRule.setAssignment>>
-            <br>$args[0].slaveName is now allowed to select her own assignments.
-        <</if>>
-    <<elseif ($args[0].assignment !== _currentRule.setAssignment)>>
-        <<assignJob $args[0] _currentRule.setAssignment>>
-        <br>$args[0].slaveName has been automatically assigned to _currentRule.setAssignment.
-    <</if>>
-    <</if>>
+	<<if (_combinedRule.setAssignment !== "none")>>
```
Frodfurgurn1991: Please use `_currectRule` for all these still, using `_combinedRule` would just lead to confusion down the line, since it is indeed the current rule that is processed in this loop still.

src/utility/raWidgets.tw
```diff
@@ -3281,210 +3294,160 @@ Your brand design is ''$brandDesign.''
...
-/% In case a slave has no rules applying to them, _combinedRule would (should)
-be an empty object, and that would wreck havoc to the rest of the code.  So, we
-check if a rule attribute, any one, is defined and otherwise skip everything.
-%/
-<<if (def _combinedRule.clothes)>>
-
 <<if $args[0].fuckdoll == 0>>
 
-<<if (_combinedRule.clothes !== "no default setting")>>
+<<if (def _combinedRule.clothes) && (_combinedRule.clothes !== "no default setting")>>
```
Frodfurgurn1991: As mention in the PR already, this is not needed, the `<<if>>` above is enough... trust me on this 😛

Frodfurgurn1991: same with all the following similar changes

fcpmanon:
On review of the diffs, everything looked OK to me.

Rule 1 Behavior Rules: 
**Assignments: selected by the slave. Player control** 
Living standard: no default setting. No default setting | Luxurious | Normal | Spare 

You removed that section, but you replaced it with the | **Let her choose** link in the setAssignment section.  Which makes sense.

fcpmanon:
Since this was rebased on the other ra-merge-again branch, it is no longer a direct successor to the ra-fix branch you used to create #442 so can you close that PR now to avoid conflicts?

src/utility/raWidgets.tw
fcpmanon: if you pick a facility to auto-assign slaves into, the starting assumption is that you want the rule to continue to apply to them while they are in that facility, so it doesn't "stop applying to slave X" at the next rule update and pull them back out of the facility again

those facility/excludeFacility edits are meant to prevent that kind of flapping, but users can override that (if they did intend for the rule to stop applying) by fiddling with the individual "Stop" buttons on the Apply to facilities: or Exclude facilities: line, as appropriate.  needless to say, it would be wise to avoid having facilityRemove be true if you don't allow the rule to continue to apply .. but I didn't go quite that far with enforcing logical combinations of settings, just tried to set things up initially in a sensible way

src/uncategorized/main.tw
```diff
@@ -149,7 +149,7 @@ __''MAIN MENU''__&nbsp;&nbsp;&nbsp;&nbsp;//[[Summary Options]]//
 	<<else>>
 		| //<<link "Stop applying Rules Assistant at week end">><<set $rulesAssistantAuto = 0>><<goto "Main">><</link>>//
 	<</if>>
-	| //<<link "Activate Rules Assistant">><<for _i = 0;_i < _SL;_i++>><<if $slaves[_i].useRulesAssistant == 1 && $slaves[_i].assignmentVisible == 1>><<DefaultRules $slaves[_i]>><</if>><</for>><<goto "Main">><</link>>//
+	| //<<link "Re-apply Rules Assistant now (this will only check slaves in the Penthouse)">><<for _i = 0;_i < _SL;_i++>><<if $slaves[_i].assignmentVisible == 1 && $slaves[_i].useRulesAssistant == 1>><<CheckAutoRulesActivate $slaves[_i]>><<DefaultRules $slaves[_i]>><</if>><</for>><<goto "Main">><</link>>//
```
Frodfurgurn1991: The `.useRulesAssistant` check should probably go to `<<CheckAutoRulesActivate>>`

src/utility/assignWidgets.tw
```diff
@@ -36,16 +36,19 @@
...
 		<<case "choose her own job">>
 			<<set $args[0].assignment = $args[1],                       $args[0].choosesOwnAssignment = 1>>  /* removeJob already set assignmentVisible = 1 */
 		<<default>>
 			<<set $args[0].assignment = $args[1]>>                      /* removeJob already set assignmentVisible = 1 and choosesOwnAssignment = 0 */
 	<</switch>>
 
-	<<if _wID == $personalAttention>>
+	<<if _wID == $personalAttention && $args[0].assignmentVisible == 0>>
 		<<set $personalAttention = "business">>
+		$args[0].slaveName no longer has your personal attention; you plan to focus on $personalAttention.
```
Frodfurgurn1991: Are you sure this should be here?  `<<assignJob>>` seems to be called from many places and I don't know if it should be displaying this in every case.

Frodfurgurn1991:
Could you go over the comments that I left?  I don't know why they are not displaying here.

src/utility/assignWidgets.tw
fcpmanon: It worked OK when I play tested it a bit, but we'll have to see.  It displayed briefly (as a transitional black screen with just that text) when the slave being reassigned had personal attention.

src/utility/raWidgets.tw
Frodfurgurn1991: Well, previously that `_exclude` was at the end of the loop so only the assignments were skipped, like I did.

src/uncategorized/main.tw
fcpmanon: Do we actually want to clear the .currentRules list of a slave that is RA-Exempt?  That would have the side effect of exercising facility remove rules, etc, unless we also add the useRulesAssistant check to DefaultRules.

fcpmanon:
I think I've responded to all comments now.  Did I miss any?

Frodfurgurn1991:
Yes, there seem to be some that are tagged as outdated.

src/utility/raWidgets.tw
Frodfurgurn1991: Well, this is not code that I know that well so I'll trust your judgement.

src/utility/raWidgets.tw
fcpmanon: OK, I'll revert that.

src/utility/raWidgets.tw
fcpmanon: OK.

src/utility/raWidgets.tw
fcpmanon: Huh.  Shouldn't we actually close the <<for>> loop so we finish merging all the rules before we start processing them?

fcpmanon: For some reason I was reading it as if that was already the case.

fcpmanon: Never mind - I think I understand what you were doing here now; the assignments portion of each rule is being applied sequentially, but the other rules are handled as merged.  That makes sense.

src/uncategorized/main.tw
Frodfurgurn1991: Are you sure that slaves are removed from their facilities when a rule stops applying to them?  I mean, the only place that reference `<<RAFacilityRemove>>` (and Rules Facility Remove) before it, is inside `<<DefaultRules>>` and it only calls it in the `<<switch _currentRule.assignFacility>>`.  In other words, for a rule to remove a slave from a facility it first has to apply to them.

src/utility/raWidgets.tw
Frodfurgurn1991: Yes, while `mergeRules()` takes extra care to probably figure out the resulting assignment or facility the slave should go in, I'd still not rely on that, at least for now.

Actually, the biggest reason that I kept all that in the loop was because, previously, special slaves were only potentially excluded from assignments, and `mergeRules()` doesn't take a slave object, so it can't know if the rules that it is merging should be for a special slave or not.  However, you now exclude special slaves from the whole of the rule, so that point is moot.  Still, if it works, don't fix it.

fcpmanon: I also just did a search for _currentRule in other files and all of them initialize it before use, so I think we're safe (for now).

src/utility/raWidgets.tw
Frodfurgurn1991: Actually, I'm ok with keeping the checks for now, mostly because of old save compatibility and stuff.  Perhaps I'll remove it in the future, but you should keep it for now.  Do also keep the other `<<if>>` though.

Sorry about that.

src/utility/raWidgets.tw
Frodfurgurn1991: But still, if the code is in the loop, please use `_currentRule`.  It's confusing otherwise.

src/utility/raWidgets.tw
fcpmanon: Even so, I think that interpretation of excludeSpecialSlaves is non-intuitive considering its current placement in the UI and how every other toggle is treated as part of the condition for the entire rule.  Do you have rules that depend on excludeSpecialSlaves only applying to reassignment?  Should we change its location in the UI so the toggle appears only if setAssignment or assignFacility is used and indented below them in that case?

src/utility/raWidgets.tw
fcpmanon: I also did a search for _currentRule in other files and all of them initialize it before use, so I think we're safe (for now).

Frodfurgurn1991: That's good to hear, but still, like I said, for a semantic reason, do use `_currentRule`, pretty please.  Same with the following comment.

src/utility/raWidgets.tw
Frodfurgurn1991: Well, I personally don't assign slaves through rules so I don't care much for it.  In the end, I think we should just ping @freecitiesdev for his opinion on it.

src/uncategorized/main.tw
fcpmanon: Right.  The scenario I'm considering is that a slave was assigned to a facility by a rule (currently applied), and then you exempt that slave from the RA.  Now the slave will no longer qualify for the rule, so <<CheckAutoRulesAssistant>> will remove it.  At that point, <<RARemoveRule>>, which will call RAFacilityRemove, and if the rule had facilityRemove == true, the slave will be removed.  I'm assuming you were thinking that CheckAutoRulesAssistant would treat .useRulesAssistant == 0 as a reason to consider all rules non-applicable, rather than just turning into a no-op.  If we made it a no-op, the exempt slave's .currentRules would remain untouched but then <<DefaultRules>> would need the same check added also or else it would continue to apply whatever rules the slave had in .currentRules before setting .useRulesAssistant = 0.

src/utility/raWidgets.tw
fcpmanon: Erk, I just removed them all...

fcpmanon:
OK, I think this is now safe to merge.  The only issue not addressed is the idea of pulling .useRulesAssistant checks into <<CheckAutoRulesAssistant>>, which is a redesign rather than a fix (existing code works).

src/uncategorized/main.tw
Frodfurgurn1991: I think you meant `<<CheckAutoRulesActivate>>` there.

Well, we could special case it, so that if a rule does not apply anymore because `.useRulesAssistant` is false, then the slave is not removed from the facility.  Or we could not do that and just state that exempting a slave from the RA may remove her from any facilities she's assigned.

Or we could just not do anything 😛

fcpmanon: Yes, typo.  Since I don't know of any bugs with how RA-exemption works at the moment, I will leave it alone for now.

src/utility/raWidgets.tw
fcpmanon: I did.

src/utility/raWidgets.tw
fcpmanon: Done.

src/utility/raWidgets.tw
fcpmanon: Might be better to enforce mutual exclusion of conditions that cause flapping depending on the facilityRemove setting, but maybe some player _wants_ flapping.  We'll see.

Frodfurgurn1991:
Oh, also, if this PR fixes any of the issues, which I think it does some, you can put the number of the issue in the commit message and I think it will autoclose when the PR is pulled.  So like, write the commit and the then put a "Closes #410" in a different line there.  You can use an interactive rebase to amend commit messages.

freecitiesdev:
Questions about the current status of backwards compatibility:

- Will firing backwards compatibility manually allow continued play on an old save? Has this been thoroughly tested?

- Will using NGP allow continued play on an old save? Has this been thoroughly tested?

Frodfurgurn1991:
Well, I can only speak for the RA changes, but they should be fully backwards compatible.  Indeed, this PR only removes some variables that were not used anyway.  The only different behavior would be the way special slaves are handled.  Before, everything other than assignments would be applied to them.  After this change, they are completely exempt from any rule that mentions as such.

fcpmanon:
There's no issue taking a save on the current master commit (pre-PR) to a game with the PR applied.  Older saves should be handled correctly by the backwards compat update (manually run via options menu link); I've tested that in the past, and it hasn't been changed by this PR.  NGP uses the backwards compat automatically, so should also be fine.  Also I tested the FCDev's "preferred" rules just now with PR applied and all was well.  However, you might want to update those, as they don't have the new smart clit piercing variables in them.

Issue #417, opened 2017-05-07
Collars are not marked harsh/nice in unique slave setting: compare to clothing.


lowercase-donkey:
I assume it's:

Harsh collars:
`Tight steel | Cruel retirement counter | Uncomfortable leather | Shock punishment | Dildo gag | Neck corset` 

Nice collars:
`Stylish leather | Satin choker | Heavy gold | Pretty jewelry | Nice retirement counter | Leather with cowbell | FS Bowtie collar | FS Ancient Egyptian `

But I wasn't sure about neck corset.

Issue #409, opened 2017-05-06
Divergence between RA and unique slaves: Assignments/facilities


lowercase-donkey:
Now with tables!  Tables just for git, it's easier to compare several lines this way.

Assignments:

| RA: Apply to assignments: All | Rest | Fucktoy | Subordinate Slave | House Servant | Confined | Whore | Public Servant | Classes | Milking | Gloryhole |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- 
| Unique: Assignment: | Rest | Fucktoy | Subordinate slave | Servant | Take slave classes | Confine her | Whore | Public servant | Glory hole | Get milked | Let her choose her own assignment | Assign her to her current task for another month 

Facilities:

| RA: Apply to facilities: | All | The Head Girl Suite | The Brothel | The Club | The Arcade | The Dairy | The Servants' Quarters | The Master Suite | The Schoolroom | The Spa | The Clinic | The Cellblock 
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- 
Unique: Transfer to: | Brothel | Club | Dairy | Servants' Quarters | Master Suite | Spa | Clinic | Schoolroom | Cellblock | Arcade 

As you can clearly see, not only does order diverge between the two pages, but naming does as well.

Let me temporarily correct order so it is easier to compare naming:

| RA: Apply to assignments: All | Rest | Fucktoy | Subordinate Slave | House Servant | Confined | Whore | Public Servant | Classes | Milking | Gloryhole |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- 
| Unique: Assignment: | Rest | Fucktoy | Subordinate slave | Servant | Confine her | Whore | Public servant | Take slave classes | Get milked | Glory hole |  Let her choose her own assignment | Assign her to her current task for another month 

I simply had to move classes down, and switch milking/glory hole.  Should be simple to standardize naming (House servant vs servant, Gloryhole vs "Glory hole" etc.)

Facilities is a bit of a mess: I moved arcade and spa

| RA: Apply to facilities: | All | The Head Girl Suite | The Brothel | The Club | The Arcade | The Dairy | The Servants' Quarters | The Master Suite | The Schoolroom | The Spa | The Clinic | The Cellblock 
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- 
[...]

Issue #385, opened 2017-05-05
Sorting clothing alphabetically.


lowercase-donkey:
I'm still trying to understand clothing, so bear with me.  If the "slutty/nice" modifier is moved to the end of clothing titles, I think the ordering would be:

```
Harsh:

Uncomfortable straps
Shibari ropes
Restrictive latex
Penitent nun


Nice:

Bangles
Bodysuit
Cheerleader outfit
Clubslut netting
Cutoffs and a t-shirt
Fallen nun
Haltertop Dress
Hijab and abaya
Latex catsuit
Leotard
Maid (Nice)
Maid (Slutty)
Military Uniform
Mini Dress
Nice lingerie
Nurse (Nice)
Nurse (Slutty)
Schoolgirl
Silken Ballgown
Skimpy Battledress
Slave Gown
String bikini
Succubus Costume
Suit (Nice)
Suit (Slutty)

```

I'm having a hard time piecing in FS and purchased items, and somethings are named a bit inconsistently.  (Is "Slutty jewelry" = Bangles?)

I think FS is:
```
FS Chattel habit
FS Conservative clothing
FS Japanese clothing
FS Western clothing

```
But I know I am missing some.  Also not included were some items that could be bought.  Is there crossover between FS and directly purchased?  How are they labeled when choosing clothing if that is the case?  Anyway, insert as needed:

```
body oil
bunny outfit
chains
slutty qipao
toga
```

lowercase-donkey:
```
Harsh clothes: 	
	
Penitent nun		penitent nuns habit
Restrictive latex	restrictive latex
Shibari ropes		shibari ropes
Uncomfortable straps	uncomfortable straps
	
Nice clothes: 	
	
Bangles			slutty jewelry
Bodysuit			comfortable bodysuit
Cheerleader outfit	cheerleader outfit
Clubslut netting	clubslut netting
Cutoffs and a t-shirt	cutoffs and t-shirt
Fallen nun		fallen nuns habit
Haltertop Dress	halter top dress
Hijab and abaya	harem gauze
Latex catsuit		latex catsuit
Leotard			leotard
Maid (Nice)		nice maid outfit
Maid (Slutty)		slutty maid outfit
Military Uniform	military uniform
Mini Dress		mini dress
Nice lingerie		attractive lingerie
Nurse (Nice)		nice nurse outfit
Nurse (Slutty)		slutty nurse outfit
Schoolgirl		schoolgirl outfit
[...]

Issue #384, opened 2017-05-05
Divergence between RA and unique slave settings, some minor


lowercase-donkey:
Typical punishment/reward is on the same line in unique slaves, two lines in RA.

Living standard options are in reverse order between unique and RA.

"Non-assignment orgasm rules" vs "release rules"

"Go naked" vs "nude"

Harsh clothes vs Nice clothes: categories reversed in order, labeled differently (sometimes parenthesis for "nice/harsh," sometimes not) and are scrambled in terms of items.  I understand wanting to keep *nice/slutty* items together, but I don't understand the ordering of the rest of the list.  Perhaps change variants to `nurse (slutty)` and then alphabetical?  While keeping FS at the end? #385

Collar: just easier to put them all here, pretty wild divergence:

Unique slaves:
`None | Tight steel | Cruel retirement counter | Uncomfortable leather | Shock punishment | Dildo gag | Neck corset | Stylish leather | Satin choker | Heavy gold | Pretty jewelry | Nice retirement counter | Leather with cowbell`

RA, plus "No default collar" setting, of course
`No collar | Steel | Cruel retirement | Harsh leather | Shock | Dildo gag | Neck corset | Nice leather | Satin choker | Gold | Pretty | Nice retirement | Cowbell `

Shoes too:

`Go barefoot | Flats | Heels | Thigh boots | Painfully extreme heels`

`Barefoot | Flats | Heels | Boots | Extreme heels `

"Body Accessory" vs  #"Corsetage." #382

Chastity see #379

Genuinely confused about "drugs" vs "growth regimen for healthy slaves."

"Health" vs "Health drugs"

Aphrodisiacs is on the same line as health in Unique, but not RA

"Diet" vs "Muscles," although I understand they are different things they seem similar.

fcpmanon:
Body Accessory changed to Torso Accessory in #443

fcpmanon:
Open issue - muscles setting alone has no effect if diet is "no default setting"

PR #447, opened 2017-05-12
New to project and git, hope this helps

lowercase-donkey:
Finally diving in! I know some of the sorting is silly to get into now, but I need to start small and get comfortable.

These changes revolve around my goals of:

1. Intelligible sorting of lists
2. Keeping RA and slave's Interact screens as similar as possible

These things were driving me bonkers. Hope this helps the project.

I tried to address:
Sorting Jobs: #409 
Collars: #417 
Clothing: #385 
And a few of these: #384

### DIFF
--- a/src/uncategorized/rulesAssistant.tw
+++ b/src/uncategorized/rulesAssistant.tw
@@ -345,22 +345,15 @@ Clothes:
 	<<RAChangeApply>>
 <</link>>
 |
-<<link "Nice lingerie">>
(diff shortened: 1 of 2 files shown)


lowercase-donkey:
If I keep working on this, should I be making RA look like slaves or slaves like RA?

For example, slaves use underline (which doesn't look that hot) but RA does not.  Trying to get a mental style guide here.

Issue #400, opened 2017-05-06
"Neon chocolate" as a descriptor


lowercase-donkey:
```
You call out to stop her, and she turns obediently to listen; you tell her to take the day off and meet you that evening for a trip to Peak's most fashionable nightclub. You emphasize slightly that it's a place you prefer to enjoy with a young slave, and her eyes widen a little at the implied compliment and challenge. Right at the proper time, she arrives in your office wearing neon chocolate brown makeup to match her hair, and a tiny iridescent clubgirl outfit of the same color. 

```

"Neon chocolate makeup" isn't really a thing, was neon hard coded?

Psi-Mage:
Yeah, I always wondered what "neon black" looked like.

freecitiesdev:
It's hard coded and I'm not adding a color parser just to figure out what should and shouldn't be neon.

PR #453, opened 2017-05-12
Neon

lowercase-donkey:
Still learning how all this works.  Would this be a decent way to handle #400 instead?


freecitiesdev:
Looks fine. It also looks like you're stacking all of your changes into each pull request, which isn't fatally compromising anything as far as I know but produces excessive commit lists.

lowercase-donkey:
I wasn't sure how to stop doing that.  Thanks though!

PR #456, opened 2017-05-13
Improve weekly reports; fix "choose own assignment" bugs; fix more RA bugs; new feature for Brothel and Club (auto-apply preventatives); suppress some reputation gain text when reputation is capped; several other fixes (including an infinite loop on main)

fcpmanon:


fcpmanon:
no further bugs encountered yet in playtesting - I think this is ready to merge

freecitiesdev:
>suppress most "society approves ..." text in weekly reports when reputation is capped -- this significantly reduces clutter in the weekly report without loss of content, since you still see everything when it matters: while building reputation

Embarrassed I didn't think of that a long time ago.

freecitiesdev:
Fair warning: I'm about to revert the preventative in facilities stuff in an upcoming PR. Facilities were already modeled as applying preventatives, but badly. I've addressed the issue a different (and more efficient) way.

fcpmanon:
Saw that.  It's a cleaner solution; no need to worry about it.

PR #457, opened 2017-05-13
Bulk Slave Purchase

DrNoOne0:
Adds the ability to buy multiple slaves at once. 
* Cascading purchase allows you to inspect and buy multiple slaves before introducing them.
* Bulk buy allows you to buy 5 slaves from the same source at the same time, you will not be able to inspect them before the purchase and you'll get a 5% cost reduction. 
* The bulk buy option only shows up when you have enough to buy 5 minimum cost slaves (based upon policies, not the market's actual cost), and will cut you off if you cannot afford all five slaves.
* Bulk bought slaves will count towards a Slave School's slave count, but will not interact with the half-price sale at a school.

Behind the scenes notes:
* Pushes all the slave market slave generation code into a new GenerateMarketSlave widget, located in slaveCreationWidgets.tw.  
* Utilizes craftiness so that the multi slave intro uses the single slave intro code.


freecitiesdev:
I take it this has been thoroughly tested? If this goes wrong, we'll get into nuke-it-from-orbit territory, fast.

DrNoOne0:
Yeah, it has been thoroughly tested.  Some of the messages could be improved (primarily in the Bulk Slave Intro). 

 I didn't test with all the SMRs, but I know Mandatory Bimbofication and Universal Lactation both work, so the others should.

PR #471, opened 2017-05-16
fix syntax error in ghanan names

goldifrocks:


fcpmanon:
thank you, thank you, thank you

PR #476, opened 2017-05-17
Fix duplication of bulk purchase slaves (caused by unintended interaction with gingering); fix gingering and law compliance interaction (gingering overwrote law compliance effects); fixes for Aztec FS; other incidental fixes; reinstate "spots" counters for club, brothel, dairy and servants' quarters; use enunciate widgets in RESS and industrialDairyAssignmentsScene

fcpmanon:


fcpmanon:
added slave purchase fixes:  fix law compliance (SMR effects) and gingering (in general - changes to slaves' stats due to SMRs will no longer be discarded without effect if the slave is gingered) and especially for bulk slave purchase (solves duplication of 5th slave, which was caused by gingering code reverting $activeSlave to the last one that went through Law Compliance)

note that bulk slave purchase will apply stats changes and gingering to obtain an "estimated" price, but then $activeSlave will be reverted and the law compliance + gingering will be applied again when displaying the slave in bulk slave intro, so the price paid for a slave in a bulk purchase will not perfectly reflect the slave's final stats because law compliance stat effects are randomized

also fix some incidental description widget text

Folant:
>leat$possessive 
>Schoolteac$possessive

Beautiful.

PR #490, opened 2017-05-20
wide-ranging fixes and refactoring - includes #492 (no conflict with #491, #493)

fcpmanon:
**This includes an important reversal of setting $i to $slaves.length in AddSlave which overwrote $i in various multi-acquisition loops preventing acquisition of more than one slave from those events.**

1. tested starting new game without errors, buying lots of slaves, putting them into facilities or choose-own-job
2. tested weekly facility reports all run without errors when unstaffed or staffed with leaders and slaves inside
3. tested peLonelyBodyguard event: all actions worked as expected
4. tested PESS loving headgirl, worried headgirl, bodyguard bedtime, loving concubine and headgirl dickgirl: all actions worked as expected
5. tested reFullBed: all actions checked
6. tested pCoupAttempt: all actions worked as expected
7. tested fAbuse, fFeelings, fKiss, fRelation, fRival, fPCImpreg, fImpreg, fImpregConsummate: all worked
8. tested Remove activeSlave: worked as expected
9. tested specialSlave and completeCatalog: worked as expected
10. tested reRelativeRecruiter: worked as expected
11. tested peHeadgirlConcubine: worked as expected

I think this is safe to merge now.

changes (highlighted those with significant player-visible effect):

**do not change $i in AddSlave**; rename temporary variables that are used in included passages to minimize naming collisions that cause unexpected bugs in calling passages; replace various for loops with native JS array.find and array.findIndex for improved performance and reduced chance of errors (mostly focusing on tw files for events and tw files that are included by others, facility weekly reports among them); rework PartnerVCheck in miscWidgets; **artWidgets fix for prosthetics reported by anonymous blog poster**; cleanup nobr in various passages; **fix errors with shelter slaves**; various event fixes in PESS.tw; fix incorrectly setting arcology government property to an integer (should be a string) in various files; fix randomization of government types for starting arcologies; fix randomization of directions for arcologies added by cheat editing; **fix errors with traitor recovery in pCoupAttempt**; change encyclopedia link when viewing penthouse; update leader variables in asDump; **fix slaves self-assigning to master suite**; **fix corporate market being available to build in markets sectors before incorporation**; **fix reversed useTabs option checks**; **fix neighbor arcology numbering**; **fix missing _partnerSlave default assignment text in WalkPast**; improve cheatMode event selection to allow more back-and-forth navigation while testing; encode all touched files in UTF-8

PR #493, opened 2017-05-21
Pit decomm. bug fix

Folant:


Folant:
Oops, it seems saLongTermEffects was fixed in #490 ( https://github.com/Free-Cities/Free-Cities/pull/490/files#diff-5862cec86a182b34f5f3a1e1e5f31fe6 ).

PR #494, opened 2017-05-22
Match order of jobs in slave summary

lowercase-donkey:
Now consistent with both RA and slave interact.

Does explicitly saying "Improvement/Work" here seem to confuse things? I shortened it a bit, but it could be dropped completely.

As in other places, had to change Fuckdoll check to check every assignment individually, instead of once for a block.  I don't know if it slows things down, but it allows the order to be changed freely.


lowercase-donkey:
Also as a thought?  Slave name is of variable length.  At the cost of vertical space, we could move assignments to a new line, and preserve the left/right location of assignments on the screen.

freecitiesdev:
Can this conflict be addressed, or should I close this?

lowercase-donkey:
I'll try to fix.

freecitiesdev:
I apologize for the trouble your PRs seem to be running into. You're doing good work, it's just that this kind of cleanup stuff tends to fall foul of other changes.

lowercase-donkey:
Thanks!  Was really hoping it helped, and that's very very encouraging to hear.  Hopefully (thanks again to suggestions) by doing a script you can fire whenever you have a clear shot, this should be easier.

Please let me know if there's any tweaks that would help.

freecitiesdev:
> Please let me know if there's any tweaks that would help.

Your cleanups have been great so far. If you play FC and correct whatever offends your eye, that'll be a great help indeed.

PR #496, opened 2017-05-22
Reorganize files

lowercase-donkey:
Created new categories when appropriate.  Time consuming, but hopefully helpful.


lowercase-donkey:
Looks like it automatically stripped out the empty "uncategorized" folder, apologies.  It's a good folder to keep around.

fcpmanon:
I had a similar idea but was less ambitious in scope and then backed it all out of my pending merge request because it made it harder to see everything else I did

lowercase-donkey:
Sorry for dupes, I think GitHub for windows derped when I switched branches, and then back.  Used [this guide](http://stackoverflow.com/questions/5783159/checking-for-duplicate-file-contents-in-git) and it looks fine now.

fcpmanon:
Perhaps instead of doing this as a pull request, you could make a script with all the move/rename commands in it, and FCDev could run that once and then commit the changes.  It would be much easier to review that way.

lowercase-donkey:
It's bonkers to make though.  Could I get a pull in waves, starting with less common files?  Trying to do this by hand is almost impossible.  The merge would be effortless if could be timed.

PR #503, opened 2017-05-23
futureSocieties.tw fix

Folant:
100k as upper bound is probably enough


freecitiesdev:
10k/week cap is intended.

PR #513, opened 2017-05-24
major bug - fix rival arcology state never clearing due to wrong variable check; also hide+disable next button for certain key events until user decision

fcpmanon:
hide+disable next button for certain key events until user decision (new UpdateNextButton widget to miscWidgets and changes in storyCaption to support this); fix bug with rival arcology state never being cleared in pRivalryVictory / pRivalryActions due to checking wrong variable ($rivalrySet instead of correct $rivalSet)

tested that all events I modified display without errors and all actions can be selected and continue button appears as expected

rationale for not allowing skip during pDefenseFears and pRivalryVictory (all others still allow skip):

pDefenseFears - skipping with continue avoids the intended negative consequence of refusal (rep and prosperity penalty)
pRivalryVictory - failure to select an action results in rival arcology state never clearing


fcpmanon:
checking conflict

PR #517, opened 2017-05-25
slaveSummary.tw fix for 0 musc.

Folant:
That whole ````'if $summaryStats~'```` thing throughout slaveSummary.tw really suck, but I haven't found the way to save colors without them.


fcpmanon:
Flabby?  Why not "soft" or something less opinionated?

Folant:
I'm bad at choosing words

PR #520, opened 2017-05-25
slaveSummary.tw rework

Folant:
Only ~~30~~ 26 ````<<if $summaryStats```` now vs 138 previously  
codedoge 1


freecitiesdev:
Oh God no.

PR #533, opened 2017-05-26
add missing research lab link to buildingWidgets; fix short-circuit bug in genericPlotEvents caused by nextLink being erroneously set to Main when acquiring a slave, thereby bypassing week advancement process; redirect nextLink from wardrobe back to penthouse instead of Main; fix head girl changing fetish of mindbroken slaves in HG suite (thereby restoring them); fix head girl abuse of barely-devoted slaves in HG suite (devotion = 50) due to off-by-one comparisons

fcpmanon:


freecitiesdev:
Great work.

PR #539, opened 2017-05-28
gingering and acquisition fixes

fcpmanon:
fixed gingering state carrying over to different slaves when a gingered slave is inspected but not purchased (slaveCreationWidgets + newSlaveIntro); also changed New Slave Intro to set nextLink to AS Dump if not already set (New Slave Intro will first set returnTo to nextLink to preserve original passage sequence - this ensures that New Slave Intro's changes to slave stats will not be lost, as was happening in several events that called AddSlave first - however this breaks Bulk Slave Intro's dependency on $returnTo being unchanged, so New Slave Intro will check passage() and avoid setting nextLink/returnTo if it is Bulk Slave Intro as a special case); no changes to Bulk Slave Intro itself; cleaned up various occurrences of (now unnecessary) setting nextLink to AS Dump / calling AddSlave / including New Slave Intro for consistency and to avoid losing New Slave Intro stat changes; corrected dangerous nextLink = "Next Week" in reMilfTourist (which appears to be an unused event); added New Slave Intro to RE Royal Blood event for single-slave acquisitions only; added New Slave Intro to P slave medic event; also corrected some misuse of "Remove activeSlave" (RES Failure - did not set $activeSlave before removing) (reFSAcquisition - the slave being removed hadn't been added yet); fixed a couple of places where scheduled events were not looping back to scheduled event again, so if you have a recruitment success plus special slave order in the same week, for example, you will now see both results in the same week

### DIFF
--- a/src/npc/removeActiveSlave.tw
+++ b/src/npc/removeActiveSlave.tw
@@ -1,4 +1,4 @@
-:: Remove activeSlave [silently]
+:: Remove activeSlave [silently nobr]
 
 <<set _ID = $activeSlave.ID, _SL = $slaves.length>>
 
--- a/src/npc/removeActiveSlave.tw
+++ b/src/npc/removeActiveSlave.tw
@@ -39,7 +39,10 @@
 	/% Remove from facility array or leadership role, if needed %/
 	<<removeJob $activeSlave $activeSlave.assignment>>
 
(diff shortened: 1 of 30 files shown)


freecitiesdev:
You've done more work on FC than I have recently. And that's not something I say frequently, or lightly.

PR #540, opened 2017-05-29
increase HG skill training bonuses when HG has complementary traits (was 1-2, now 5-10)

fcpmanon:


lowercase-donkey:
I'm kind of curious about how the madam works.  Does she improve slaves the same way as HG?  If so, does she do it with the same limitations?

fcpmanon:
Madam affects brothel profits but does not affect whores' rate of skill improvement

fcpmanon:
Also I misrepresented the HG rate of skill gain.  A brilliant HG over 35 working with a brilliant slave could improve a single skill by as much as 17 points per week using the original code.  With this pull request, that potentially goes up by a 10-20 points, so as much as 37 points per week.  Still three weeks to max a single stat for a brilliant with continuous focused attention from a max-skill brilliant HG.  For comparison, a brilliant slave working in the brothel can gain 12+ skill points per week.  I think HG training should be better than just working in a facility.  Note that HG training can _stack_ with working in a facility if universal rules are set to "cooperate with facilities" and there is free space in the facility and the slave is assigned in the penthouse to the equivalent job (whore/serve the public).  Of course all of these point values go down with lower intelligence.

freecitiesdev:
Since you're hip-deep in the code, I'm willing to have you be aggressive with the balance. We can always rebalance if it's too much.

PR #556, opened 2017-05-30
CRITICAL: fix mismatched closing tags in RETS to prevent infinite loop; MINOR: add modest clothes to backwards compatibility - move to "setup" global object for better performance ($ vars that never change should move to setup); also fix two wrong index variables in backwards compatibility (_i instead of _bci)

fcpmanon:


freecitiesdev:
This is getting ridiculous. I step away from RETS for one second to figure out why the vector art is misbehaving and you've already fixed my bullshit. Great work.

PR #560, opened 2017-05-30
activeSlave.recruiter bugfixes

tc-k:
activeSlave.recruiter bugfixes for consistency.

Previously the coders seem to waffle on what $activeSlave.recruiter == young sister means. Sometimes it's treated as if the activeSlave is the younger sister, sometimes it's treated as if the recruited slave is the younger sister. Now the code should be consistent. If $activeSlave.recruiter is young sister, then it is the $activeSlave who is the younger sister. If $activeSlave.recruiter is older sister, then $activeSlave is the older sister.

Also, in newSlaveIntro, moved the check for age > 18 to older sister. It doesn't make sense for the age check to happen for the young sister value.

### DIFF
--- a/src/uncategorized/newSlaveIntro.tw
+++ b/src/uncategorized/newSlaveIntro.tw
@@ -100,7 +100,7 @@ The legalities completed, ''__@@.pink;$activeSlave.slaveName@@__'' <<if ($active
 	<<set $activeSlave.recruiter = "daughter">>
 <<elseif ($activeSlave.age < 43) && (random(1,100) <= 20)>>
 	<<set $activeSlave.recruiter = "older sister">>
-<<elseif ($activeSlave.age < 25) && ($activeSlave.age > 18) && (random(1,100) <= 20)>>
+<<elseif ($activeSlave.age < 25) && (random(1,100) <= 20)>>
 	<<set $activeSlave.recruiter = "young sister">>
 <</if>>
 <</if>>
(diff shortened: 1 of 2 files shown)


lowercase-donkey:
> Also, in newSlaveIntro, moved the check for age > 18 to older sister. It doesn't make sense for the age check to happen for the young sister value.

On the contrary, since 18 is the minimum, shouldn't the smaller value (younger sister) be checked against it?

tc-k:
1. You're right about the inconsistency part. (scratches head) I'm not sure how that happened. I blame lack of sleep. Oh, wait, now I remember. I think I was noticing that when $activeSlave.recruiter is mother, the code recruits her daughter and vice versa (i.e., the .recruiter value is the target) while if it's older sister, it recruits the slave's older sister (i.e., the .recruiter value is the recruiter). Add the lack of sleep and I misunderstood the code. I could've sworn that 
2. Ignore the modification I made for the age check, as that's wrong now. However, the age > 18 check is definitely odd. For example:
If $activeSlave.age is 18-42, there is a chance to recruit her older sister, who is going to be 2 years older. So in theory the older sister is 20-44 years old.
If $activeSlave.age is 19-24, there is a chance to recruit her younger sister, who is going to be 2 years younger. So in theory the younger sister is 17-22 years old. The possibility for 17 alone is odd. And then there's the question why the possible ages are so different. What, if a slave is 35 she may have an older sister but can't possibly have a younger sister?

I'll leave this up for a bit before closing it.

PR #576, opened 2017-06-01
Salon spaces

lowercase-donkey:
Why does this matter?  If a phrase is after
`<br>&nbsp;&nbsp;&nbsp;&nbsp;` on the same line, it is proceeded by four
spaces, as expected.

However, moving the following text to the next line causes it to be
proceeded by FIVE spaces, making the indents ragged.

The quick fix was to do a line break after every "indent," giving us a consistent (five) spaces, but is there something else going on here?


freecitiesdev:
> The quick fix was to do a line break after every "indent," giving us a consistent (five) spaces, but is there something else going on here?

There shouldn't be. FWIW, I've been trying to standardize on the five space tab as I rework old code.

PR #577, opened 2017-06-01
Fix salon file

lowercase-donkey:


lowercase-donkey:
For some reason, git locally was saying that all of the content of salon was being replaced, but that mistake is not happening on web, thankfully.

The main thing here is that there seems to be a duplicate section in salon that's actually showing duplicate text sections for players.

PR #579, opened 2017-06-01
Don't merge: I have a question

lowercase-donkey:
My recent work on salon was very confusing.  Unlike other files I have been working on, GitHub for windows always showed any change as "You deleted all the text, and replaced all the text."

I've also realized that if I move a *.tw file from one folder, commit, then move it back, and commit again, many of the line ends change (red ^M.)  Because I use windows, perhaps?

However, once this happens, I can edit the effected files and my changes show normally in Git for Windows.

This commit is what happens if you move "uncategorized" into a temp folder and then out again.  It touches MOST files in that folder, but not every file?  Help a guy out?  I'd like to be able to edit things like salon.tw and see my changes previewed locally (as happens with other files.)    I'm using notepad++ for edits, should I set eol to something other than windows?

With the exception of this anti-commit, salon edits looked like they should on the website too, so there is a split between Git for Windows and the web.


fcpmanon:
I think this is an unintended consequence of the new .gitattributes file I added in my last merge request.

The "you changed all lines" diffs you're seeing are usually caused by different line endings (Linux uses LF only, and this project stores the files online with LF only; Windows uses CR LF).  When you see ^M at the end of lines in the web-comparison, those are CRs.

Normally when you check out files, git takes care of converting the line endings from LF to CRLF for you, if it recognizes the files as text, but the .gitattributes file controls exceptions to that behavior.  I added "eol=crlf" to the .gitattributes file, which is causing git to always put CRs on the ends of the lines when you check out files, and I guess it is not converting them back to LF when checking in.

I'll open a pull request to correct the .gitattributes file.

Incidentally when you discover that a pull request had unintended consequences, it may be safest to just revert it (completely undo the changes) instead of trying to manually fix whatever went wrong.

PR #605, opened 2017-06-01
CRITICAL: completely remove CRLF end-of-line setting from gitattributes

fcpmanon:
Seems to be causing severe problems for other committers.


lowercase-donkey:
Thanks!

freecitiesdev:
Can't tell you how relieved I am that there's an identifiable fix for this. I was rather concerned last night.

PR #628, opened 2017-06-02
fix RA applying corrective glasses/contacts to blind slaves; add line break between neighbor development paragraphs in weekly report; fix events that have an "exit clause" (goto a different passage if conditions are not satisfied) but had code that still executed afterward; fix logic errors in reBusyMasterSuite event (anal virgins are now handled correctly); also fix typos in pBioreactorPerfected; also update name span in reBusyArcadeWorkaround after selection; also change reLegendary events to increment prestige by 1 (up to max of 3) instead of setting to 1, in case the slave selected for the event is already prestigious; also fix misplaced line break in reNickname when first action option is hidden; also simplify code in reStaffedMorning; add a "none" option for growth settings in RA rules

fcpmanon:


fcpmanon:
Where we end up after those last two commits is:  corrective glasses only for nearsighted slaves, never for blind slaves, but still decorative glasses for blind slaves if and only if "universal glasses" is chosen in the RA

freecitiesdev:
Great work.

Minor thing, but would you mind titling PRs like this something like "minor bugfixes" and then putting the full list in the description? Some Git interfaces don't handle 100+ word PR titles well.

PR #635, opened 2017-06-03
New Event: Play poker with the mercs

freecitiesbandit:
New event involving a poker night with the mercenaries. There's a few paths and uses the random feature so it doesn't always play out the same way.

### DIFF
--- a/src/uncategorized/PESS.tw
+++ b/src/uncategorized/PESS.tw
@@ -71,10 +71,6 @@
 
 <<switch $PESSevent>>
 
-<<case "bodyguard beatdown">>
-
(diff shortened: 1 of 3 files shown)


Boney-M:
Losing 1000 rep for not attending their poker night? That seems off, especially when the player could just hit 'continue' and not lose any.

freecitiesbandit:
The rep loss was unintentional, thanks for pointing that out.

IMO, the costs for a lot of the story events are a little too low. If I were to divide the costs in the event by ten, you only get 1000 for a 'big' payout and 100 for a 'safe' payout. At that low a reward, what would the point be? Could still use some tuning though. Maybe 2000-5000 for the big payout?

Boney-M:
5k for the buy-in and thus for the big payout would make it equivalent to the shotgun match, so that would fit well, especially since it's explicitly stated that it's the officers and not the rank-and-file you're playing with.

freecitiesbandit:
I concur and have edited the numbers to reflect that.

PR #636, opened 2017-06-03
Changed minimum rep to lose rep from not hosting parties to 1500

Boney-M:
It makes sense that nobody wants to go to a resented person's parties anyway, and it prevents players from getting stuck in a rep hole from being unable to host parties because they don't have enough rep to start hosting parties.


freecitiesdev:
Can you write a one sentence addition to the report that explains that this is happening?

Boney-M:
Done: #664

PR #640, opened 2017-06-03
Major shuffle: Sort slaveintro

lowercase-donkey:
Overall, I'm trying to make the long list easier to parse visually.

I've done this by breaking the options into two categories, focusing on modifying a slave and those focusing on interacting with a slave sexually. This way, even if some options disappear for a particular slave, a player still has a few other visual anchors to quickly find the option they want. (I myself tend to pick certain options consistently in a playthrough, depending on the character of the Master I've chosen. It's annoying to hunt through the list constantly to see where my favorite option is, particularly when the length of the list changes constantly.)

After that, I tried to sort lists from less to more severe, with very similar options moved to share a line.

I've also highlighted novel or rarely appearing "case" based options at the top of the list explicitly with an italic reference.


lowercase-donkey:
Assuming this isn't a bug, this seems to reveal to players that all purchased slaves are "migrants."  

If it's not a bug, I could conceal this by making the "migrant" entry more neutral.

lowercase-donkey:
Looks good to go, I assume the others fire in very rare cases.  It could be neat to have some more entries here for more of the re slaves someday, might have to work on that.

lowercase-donkey:
Ready for merge.

freecitiesdev:
I'm holding on to this until after the final 0.10.3 hotfix.

PR #649, opened 2017-06-03
More chastity checks and some ptWorkAround fixes.

pregAnon:
Blindness could not be applied to "comfortable seat" without a reworking, so it has been restricted accordingly.

The random event checks are painful.  You pretty much have to bring up every event to check and see if it is safe to use a canDo___ function in place of the chastity checks.

RESS changes will be pushed after the pull requests are cleaened out tonight.


freecitiesdev:
>The random event checks are painful. You pretty much have to bring up every event to check and see if it is safe to use a canDo___ function in place of the chastity checks.

Believe me, I know. In a perfect world I would have had the skills and knowledge to write a standard event design document that would regularize all these, but I didn't.

Issue #602, opened 2017-06-01
Appraisers are broken


Boney-M:
A reputable slave appraiser arrives promptly to inspect her and certify her qualities for sale. The appraiser, 

Text cuts off abruptly there.

Possibly unrelated: turning off appraisals in Game Options gives the error
 
    Error: <<set>>: bad evaluation: State.variables.seed.random is not a function

indicating the code:

    <<set $appraiser = $seed.random()>>

Boney-M:
#650

PR #650, opened 2017-06-03
fix sellSlave

goldifrocks:
the conditionals got a bit jumbled, also there are no male/female variants for "Your Greatness".


Boney-M:
Pretty sure this fixes #602

PR #651, opened 2017-06-03
bugfixes for slave intro scenes

tc-k:
There was a conditional statement ($activeSlave.origin == ) within the <<case>> that's messing the switch.

Also the case for recruiterSister would never fire since the reRelativeRecruiter.tw now specified older sister, younger sister, or twin sister, while the case is looking for just sister. So added the necessary cases so the intro will fire.


lowercase-donkey:
I found that and fixed as well in #640, so whichever gets merged first I guess.

PR #654, opened 2017-06-04
Gave St Claver's slaves lipo'd waists

Boney-M:
Spitballed it at -50, but I could see some pretty convincing arguments for anywhere between -25 to -100 for being the 'ideal' a cosmetic surgery obsessed school would shoot for. The important thing is that they're not chubby any more.


freecitiesdev:
-50 makes sense. -100 is hardcore Transformation Fetishist territory.

PR #664, opened 2017-06-05
Further clarified social event reputation changes

Boney-M:
Added text for not losing reputation because your reputation is already low and nobody wants to come to your parties anyway.


freecitiesdev:
Thanks. Since you know the code very well at this point, this is something worth keeping an eye out for: hidden mechanics that aren't reported. I'm fixing them as I find them, either by describing them as here, or removing them if they're pointless.

Boney-M:
Alright, I'll start opening issues or writing code fixing them as I find them.

PR #666, opened 2017-06-05
Fixed error in slave summary

Boney-M:
Error caused by #655 - mea culpa


Boney-M:
Downloaded the branch on my fork with that change made and it definitely fixes the problem.

freecitiesdev:
I think this may be unnecessary given #670.

Boney-M:
Yeah, #670 fixes it.

PR #676, opened 2017-06-06
Rebalanced obedience/trust fetishstrength changes

Boney-M:
1) Made it so that being sufficiently devoted and trusting isn't enough to get a slave into maximum strength and paraphilia territory - events, smart piercings, or complimentary softened flaws are needed.

2) Made it so that fetishes were a little 'stickier' on disobedient and terrified slaves - it'll still fall to a really low level, but it's less likely to disappear entirely now, making slaves a little bit more unique and stopping them all from trending to vanilla.


freecitiesdev:
I think I like these changes. Let's keep an eye on them, though.

PR #680, opened 2017-06-06
Put dick removal behind $seeExtreme

Boney-M:
Currently, with extreme content disabled, you can't remove a slave's balls, but you can remove their dick.


Boney-M:
Also tweaked the handling of nulls and the addition of vaginas to them - it's still using the mtf surgical reactions because I'm not sure what sort of reactions a slave would have to having the bare skin where a penis once was converted to a vagina.

PR #681, opened 2017-06-06
Tweaked profit reporting in dairy

Boney-M:
_profits is net instead of gross, which could lead to some misleading messages. Eliminated the ambiguity and added a message clarifying things when the dairy makes a loss - AFAIK this is only possible due to hormone implants in new cows.


freecitiesdev:
>AFAIK this is only possible due to hormone implants in new cows.

That should be correct. It used to be possible to make negative milk, but that should be fixed.

PR #685, opened 2017-06-06
misc fixes

fcpmanon:
see commits

### DIFF
--- a/src/init/setupVars.tw
+++ b/src/init/setupVars.tw
@@ -1,0 +1,8 @@
+:: setupVars [widget nobr]
+
+/*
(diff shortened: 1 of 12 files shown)


freecitiesdev:
I agree with your defining $bodyguardTrains and similarly undefined variables should be fixed when found. However, I've been doing my best to code new additions to be robust against such omissions.

fcpmanon:
It seems that what I did with the setup object initialization (moving into a widget passage) is actually the correct way to do it.  We should move more global arrays (that never change) into there.  A bonus of doing this is that we don't need to put them in backwardsCompatibility because they are always loaded.

PR #673, opened 2017-06-06
Added paraphilia satisfaction to appropriate jobs

Boney-M:
Self-neglect: Milkmaid, attendant, servant.
Dom: Stewardess,

PR #690, opened 2017-06-06
Added paraphilia satisfaction to fucktoy duty

Boney-M:
Anal addicts satisfied by being anal fucktoys
Cum addicts satisfied by being oral fucktoys for masters with dicks
Breeders satisfied by being vaginal fucktoys for masters with dicks
Self-neglectful slaves satisfied by fucktoy duty in general

As per conversation in #673


freecitiesdev:
Boney, you're doing sterling work.

PR #694, opened 2017-06-07
UPDATED AND READY TO MERGE:  move $ vars to setupVars; plus some bugfixes and refactoring in separate commits

fcpmanon:
Tested game start, bought some slaves, including shelter slave, ended week a few times, tried one variant of justice event.  Reviewed changes line-by-line in diffs, and did most of the $var to setup-var swaps via regex find-replace, so there shouldn't be too many errors introduced ... if any.

Update:  Tested coursing.  No issues caused by me, but fixed an incidental issue in assign.tw (missing goto $returnTo for assigning pit fighters and lurcher, resulting in a blank screen).  I just reviewed all the diffs again and I don't see any issues.

I think this is safe to merge now.

PR #698, opened 2017-06-07
Couple of minor fixes

Boney-M:
One a couple of instances of = instead of == caused by me, one paraphilia satisfaction check using the old cum diet code. Both pointed out by a helpful contributor via PM over on the subreddit.


fcpmanon:
I noticed the == bug and included it in #694

PR #701, opened 2017-06-08
Arcade upgrade changes

Boney-M:
Made them mutually exclusive, and the collector has a positive effect on slave health; also made the arcade report break down the results into profits from use and profits from fluids. This has the side-benefit of preventing arcade milk profits from damaging reputation - only the pure throughput on slave holes damages rep now.

Compiled and tested it from my branch, seems to be working fine.


fcpmanon:
Your commit log is hilarious.

Boney-M:
I got so close to doing the whole thing without missing an /if, too...

PR #703, opened 2017-06-08
simple solution for FSSpending problems

### LINKED ISSUE #629: Debt or General Finances Bug
An anon on /dgg/ has reported the following bug. I have no further context as to what led to this, but was asked to pass it along. 

kdt8kxk

fcpmanon:
use a temporary variable for the textbox and require user to click a separate link to save changes; also round up (Math.ceil) instead of down (Math.trunc) as it is more user-friendly that way; adjusted line breaks

this fixes #629

note that hitting Enter in the textbox no longer has any effect - this is intentional

### DIFF
--- a/src/uncategorized/futureSocities.tw
+++ b/src/uncategorized/futureSocities.tw
@@ -382,10 +382,9 @@ possible societal customizations.
 
 <<if $FSSpending > 10000>><<set _warn = 1>><</if>>
 <<set $FSSpending = Number($FSSpending) || 0>>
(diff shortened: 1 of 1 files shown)


freecitiesdev:
Should this be applied to other textboxes, and if so, why?

fcpmanon:
@freecitiesdev Possibly.  See my comment on #629.  The issue is that input to a textbox is applied immediately (as you type) to the textbox variable (the one in " "), and the input is stored as a string.  If you navigate away from the page after that without invoking some code to sanity check, you are left with a string value, which does not play well with code that expects a number.

PR #705, opened 2017-06-08
Added movie poster trinket for the 'gift' option

Boney-M:


fcpmanon:
You could also do it this way:
`<<set $trinkets.push("a poster for the movie that was made about the love between " + $activeSlave.slaveName + " and one of your mercenaries")>>`

Boney-M:
I wasn't sure if the $trinkets.push would choke on the variable, so I copied the existing code to be safe.

PR #706, opened 2017-06-08
saLongTermEffects missing bracket

Folant:


Boney-M:
_Don't be my fault, don't be my fault..._

DAMNIT

Good catch, thanks.

PR #707, opened 2017-06-08
Activation of reMilfTourist.tw

pregAnon:
It was just sitting in the code, unused, so I cleaned it up and activated it.  Might be worthwhile to consider anything else to add the the milf's slavegen, since I had only so much to build off of.

Starts showing up when arcology prosperity crests 10M and the likelyhood is determined by your reputation.  Higher = more likely, relatively speaking.


fcpmanon:
You have to set recruiter to zero AFTER New Slave Intro.  New Slave Intro has a random chance to set it.

freecitiesdev:
This is a nasty consequence of the hacked-together event system: it's not obvious if event triggers are nonfunctional or even absent. Thanks for catching this.

PR #735, opened 2017-06-10
saRelationships superfluous 'if' fix

Folant:


Boney-M:
I think that /if is closing the relationship check on line 207, and the /if issues are from my butchering of the emotional slut code. #728 rebuilds the emotional slut code and the resulting build doesn't have that error. The master branch of my fork has #728 and #730 merged into it and as far as I can tell it's free of any egregious bugs.

PR #738, opened 2017-06-10
Degredationist/Paternalist Events

freecitiesbandit:
New REFS Events

### DIFF
--- a/src/uncategorized/REFS.tw
+++ b/src/uncategorized/REFS.tw
@@ -48,13 +48,13 @@
 
 <<case "paternalist encounter">>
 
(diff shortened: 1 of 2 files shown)


lowercase-donkey:
Always great to see new content, thanks!

Boney-M:
<<if $arcologies[0].FSPaternalist > random(1,100)>>
    <<set $REFSevent.push("paternalist encounter")>>

I'd recommend setting it to something like random(20,100) so that the FS in question needs to be at least somewhat established before the event becomes possible, since the text of the event implies that the FS has achieved some traction in society. The same could be said of the other similar events.

WarlordMLP:
both the confront options have a typo. It says "that does give him impunity to ignore the law"
It should say. "but that doesn't give him impunity to break the law"

freecitiesbandit:
Just saw this. Really sorry for the inconvenience caused. Will try and be more diligent in future.

PR #741, opened 2017-06-11
Surplus /if

Boney-M:
For once, it wasn't me!


Folant:
I guess it's fixed in #694 
https://github.com/Free-Cities/Free-Cities/pull/694/files#diff-248ab9b53a2025c6f53afce413c1c96e

Boney-M:
Good catch.

PR #758, opened 2017-06-19
A bunch of small fixes

kopareigns:
An assortment of fixes for things including spelling/grammar errors, old values that were not yet updated, a small text color change, modified checks for RE devotees/rebels (and respective compatibility updates), and a QoL suggestion from reddit from the past week (also listed as [Game Mechanics](All Options Available) in the suggestions spreadsheet).

### DIFF
--- a/src/uncategorized/randomNonindividualEvent.tw
+++ b/src/uncategorized/randomNonindividualEvent.tw
@@ -202,8 +202,8 @@
 
 /* Multislave Events */
 
-<<set _devotedSlaves = $slaves.filter(function(s) { return s.devotion > 50 && canWalk(s); })>>
-<<if def _devotedSlaves[3]>>
+<<set $devotedSlaves = $slaves.filter(function(s) { return s.devotion > 50 && canWalk(s) && s.assignment != "be your agent" && s.assignment != "live with your agent"; })>>
+<<if def $devotedSlaves[3]>>
 	<<set $events.push("RE devotees")>>
 <</if>>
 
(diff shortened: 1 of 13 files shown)


src/uncategorized/randomNonindividualEvent.tw
```diff
@@ -223,22 +214,13 @@
...
-<<if ($slaves[$slaves.length-1].assignment != "be confined in the arcade")>>
-<<if ($slaves[$slaves.length-1].assignment != "work in the dairy") || ($dairyRestraintsSetting < 2)>>
-	<<set $events.push("RE rebels")>>
-<</if>>
-<</if>>
-<</if>>
-<</if>>
-<</if>>
-<</if>>
+<<if $dairyRestrainsSetting >= 2>>
```
fcpmanon: Typo Restrains -> Restraints

src/uncategorized/reRebels.tw
```diff
@@ -4,8 +4,8 @@
 
 <<set $nextButton = "Continue">>
 <<set $nextLink = "RIE Eligibility Check">>
-<<set _i = $slaves.length-2>>
-<<set _j = $slaves.length-1>>
+<<set _i = $slaves.findIndex(function(s) { return s.ID == _rebelSlaves[_rebSlaves.length - 1].ID; })>>
```
fcpmanon: _rebSlaves is undefined
Why not use _rebelSlaves[0] and [1]?  Or randomize with _rebelSlaves.pluck() on both lines.

fcpmanon: _rebelSlaves.pluck().ID

fcpmanon: On second thought, don't you need to make _rebelSlaves a $ variable since it is being set in a different passage?  Same for the devotees variable...

And then clear them in NextWeek

kopareigns: I might have to make it a $ variable. I was never quite sure what the functional difference between $ and _ variables was. When initially testing whether this still worked right, I for some reason set _rebelSlaves in both passages. After testing it again, I think you're right and I've changed it according to your recommendations (also the _devotedSlaves).

As for using length - 1 and length - 2, I used that simply for consistency with the old version. I also didn't know pluck randomized like that. Now that I'm testing pluck out, I can't seem to get it to work without adding some intermediary steps/variables. Is there a way to do it in that single line without extra nonsense? For now I've just used a temporary variable.

src/uncategorized/randomNonindividualEvent.tw
kopareigns: whoops. fixed.

PR #775, opened 2017-06-29
Various additions regarding buttplugs, prostates on females and penis enhancement

prndev:
There was a dangling if and a misplaced parentheses in the dairyReport.


src/uncategorized/saGetMilked.tw
```diff
@@ -438,6 +438,35 @@ gets milked this week.
 
 <</if>> /* CLOSES COCK MILKING */
 
+<<if $slaves[$i].genes == "XX" && $slaves[$i].prostate > 0 && $slaves[$i].balls == 0 >>
+	Her female prostate fluid is consiedered an exotic delicacy.
```
kopareigns: consiedered -> considered

src/utility/descriptionWidgets.tw
```diff
@@ -3069,7 +3069,7 @@ $pronounCap's got a
 
 <</if>> /* CLOSES DETAILED DICK DESCRIPTION */
 
-<<if ($activeSlave.dick > 0) || ($activeSlave.balls > 0)>>
```
kopareigns: Since the description starts by saying "it's not externally apparent" I'd suggest keeping these checks. 

Also, since XX slaves can be implanted with a prostate (automatic if implanted with a penis), I'm not convinced about the necessity of a check for XY genes.

prndev: I see what you mean. I did not realize that implanting a penis adds a prostate, too. At least at some point in time, the slave creation was able to come up with futanaris (women with a penis) _without_ a prostate. For them, the description is wrong: Their prostate was not removed, they never had one in the first place.

kopareigns: I understand what you're saying too. If you're worried about it being technically incorrect in case you somehow managed to get a slave with a dick and no prostate, maybe leave the if checks unchanged, and instead change the text from
`$possessive prostate has been removed`
to
`$pronoun has no prostate`

prndev: That... is actually far better.

PR #778, opened 2017-06-30
Bodymodra sorting, layout only

lowercase-donkey:
Went with the cleaner "Ears: Left | Right | Both" format instead of the ol' wall of text. Should be easier to parse and hey, now they match.

Also modRA matches autoMod now for order of tattoo areas, and they are in the same 'head to toe'  order that the brand locations from the same page have.


lowercase-donkey:
Asian art is allowed on faces in RA but not manually in autoMod. Which way should we go?

I am inclined to drop it since perma-makeup could be considered a Geisha thing.

freecitiesdev:
Yeah, drop it.

PR #782, opened 2017-07-01
Cosmetic ra

lowercase-donkey:
Change colors to match (Red vs Blazing red) move obvious dyes to end of list in RA.

Move options other than "default" to next line with indent.

Shorten default description where possible.

Change order to cut -> color -> style.  In addition to being the irl sequence, it makes clearer that cutting cuts style options.


lowercase-donkey:
Sorry about dupe with  #ab75dff, I accidentally edited my master first.  Been a while.

PR #783, opened 2017-07-01
Adding highlights to hair

lowercase-donkey:
My first time using variables: I've tried to do something similar to branding.

This doesn't exclude people from doing something silly like red hair with red highlights, etc, but I think it's a cool addition.

### DIFF
--- a/src/uncategorized/bodyModRulesAssistantSettings.tw
+++ b/src/uncategorized/bodyModRulesAssistantSettings.tw
@@ -375,213 +375,237 @@ Automatic branding is
 <span id = "brandtarget">
 Your preferred location for brands is the ''$brandTarget.''
 </span>
-<<link "Right buttock">>
-<<set $brandTarget = "right buttock">>
+<br>&nbsp;&nbsp;&nbsp;&nbsp;
+
+Ears:
+<<link "Left">>
+<<set $brandTarget = "left ear">>
 <<RAChangeBrandTarget>>
 <</link>>
 |
-<<link "Right thigh">>
-<<set $brandTarget = "right thigh">>
+<<link "Right">>
+<<set $brandTarget = "right ear">>
 <<RAChangeBrandTarget>>
 <</link>>
 |
-<<link "Right calf">>
-<<set $brandTarget = "right calf">>
+<<link "Both">>
+<<set $brandTarget = "ears">>
 <<RAChangeBrandTarget>>
 <</link>>
-|
-<<link "Right ankle">>
-<<set $brandTarget = "right ankle">>
-<<RAChangeBrandTarget>>
-<</link>>
-|
-<<link "Right foot">>
-<<set $brandTarget = "right foot">>
-<<RAChangeBrandTarget>>
-<</link>>
-|
-<<link "Left buttock">>
-<<set $brandTarget = "left buttock">>
-<<RAChangeBrandTarget>>
-<</link>>
-|
-<<link "Left thigh">>
-<<set $brandTarget = "left thigh">>
-<<RAChangeBrandTarget>>
-<</link>>
-|
-<<link "Left calf">>
-<<set $brandTarget = "left calf">>
-<<RAChangeBrandTarget>>
-<</link>>
-|
-<<link "Left ankle">>
-<<set $brandTarget = "left ankle">>
-<<RAChangeBrandTarget>>
-<</link>>
-|
-<<link "Left foot">>
-<<set $brandTarget = "left foot">>
-<<RAChangeBrandTarget>>
-<</link>>
-|
-<<link "Both buttocks">>
-<<set $brandTarget = "buttocks">>
-<<RAChangeBrandTarget>>
-<</link>>
-|
-<<link "Both thighs">>
-<<set $brandTarget = "thighs">>
-<<RAChangeBrandTarget>>
-<</link>>
-|
-<<link "Both calves">>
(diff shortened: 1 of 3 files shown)


lowercase-donkey:
I've seen conversations about local variables, is this where one should be used?  Still new to adding code, used to just shuffling it.

kopareigns:
How exactly do you want it to work? Presumably you want it to not display the string names when opening the page? I also noticed it saves the dye colors between different visits to the salon. Was this intended or did you want to reset the colors between different visits? Do you want to hide the option of dyeing her hair if you haven't selected any dyes yet?

lowercase-donkey:
I would like to reset between visits, yes.

I don't want it to display the string names, just blonde, black, etc.

I'd like to learn how to hide it with no dye selection, but I'm considering different ways to have it look.

kopareigns:
Full disclosure: I'm not any kind of programmer either so my suggestions might not be optimal.

What I would do is just add two extra variables, say $dyeSelected and $highlightSelected, that are 0 if you haven't selected a respective color, and 1 if you have. Whenever you select a dye it would change it to 1, and when you actually settle on a color and click "Dye her hair" both would be reset to 0. Around the application button you could just put an if statement that checks if both variables are 1. Suggesting a player choose a dye in an else statement is a good idea I think.

To prevent any funniness with displaying string names I would put another check where it sets both $primaryHairColor and $secondaryHairColor to "" if the corresponding variables are 0 (this could even be put into the else statement with the previous check).

So in total the area around the button might look something like
```
<<if $dyeSelected == 1 && $highlightSelected == 1>>
    [[Dye her hair|Salon][$activeSlave.hColor = ($primaryHairColor + $secondaryHairColor),$cash -= $modCost, $dyeSelected = 0, $highlightSelected = 0]] $primaryHairColor $secondaryHairColor now?
<<else>>
    Choose both a dye color and a highlight before dyeing her hair.
    <<if $dyeSelected == 0>>
        <<set $primaryHairColor = "">>
    <</if>>
    <<if $highlightSelected == 0>>
        <<$secondaryHairColor = "">>
    <</if>>
<</if>>
```
and the buttons to select a color would look something like
`[[Blonde|Salon][$primaryHairColor = "blonde", $dyeSelected = 1]]`
and similarly for $secondaryHairColor with $highlightSelected.

Finally, to reset between visits, the best way I can think of is to set $dyeSelected and $highlightSelected to 0 whenever you enter the salon from the Slave Interact passage (are there other passages you can enter the salon from?). This ensures everything is reset even if a player chooses colors and then decides to back out of the salon without applying the changes.

PS: I require both dye and highlight to be selected to prevent the weirdness that happens if you don't select a dye color and select none for the highlight. If desired this could be modified so that you can dye hair without choosing a highlight.

kopareigns:
Update: After walking away and coming back we can actually do away with both extra variables I introduced. The code is similar, we just use $primaryHairColor and $secondaryHairColor instead of the variables I introduced. There are two ways I see to do it depending on whether you 
1. Want a player to choose both a primary hair color and a highlight, or
2. Just a primary hair color is good enough. 

For the first option I might use
```
<<if $primaryHairColor != 0 && $secondaryHairColor !== 0>>
	[[Dye her hair|Salon][$activeSlave.hColor = ($primaryHairColor + $secondaryHairColor),$cash -= $modCost, $primaryHairColor = 0, $secondaryHairColor = 0]] $primaryHairColor $secondaryHairColor now?
<<else>>
	Choose both a dye color and a highlight before dyeing her hair.
<</if>>
```
and for the second option
```
<<if $primaryHairColor != 0>>
	[[Dye her hair|Salon][$activeSlave.hColor = ($primaryHairColor + $secondaryHairColor),$cash -= $modCost, $primaryHairColor = 0, $secondaryHairColor = ""]] $primaryHairColor $secondaryHairColor now?
<<else>>
	Choose a dye color before dyeing her hair.
<</if>>
```
There are some important differences. In the first notice the if check has "!==" instead of "!=". This is so that the empty string is not interpreted as 0. In this case we would want to initialize both $primaryHairColor and $secondaryHairColor to 0 when entering the salon from Slave Interact.

In the second notice we're setting $secondaryHairColor to "" when we click "Dye her hair." This is because we don't "0" as highlight. In this case we would want to initialize $primaryHairColor to 0 and $secondaryHairColor to "" when entering the salon from Slave Interact.

In both cases the buttons to select a color/highlight should remain unchanged from what you have now. Personally I think the first option is preferable since it's more clear to the player about how to change each aspect of a slave's hair.

lowercase-donkey:
How do I initialize though?

`<<set $primaryHairColor = "">>
<<set $secondaryHairColor = "">>` ?

Or initialize as 0?

kopareigns:
In slaveInteract.tw you need to change the button to bring you to the salon so that it sets those two variables. For example
`[[Use auto salon|Salon][$degradation = 0, $primaryHairColor = 0, $secondaryHairColor = 0]]`

Initializing to 0 or "" depends on the choice of implementation I mentioned above. The first choice I listed above should have both initialized to 0, and the second choice should have $primaryHairColor initialized to 0 and $secondaryHairColor initialized to "".

lowercase-donkey:
Friiiick, that's why.  That was the main thing.  You have to set it before you even hit the page, makes sense.

kopareigns:
Yeah, sorry if I went into way more detail than necessary there. Glad to help.

PR #784, opened 2017-07-02
Tailed buttplugs

freecitiesbandit:
Haven't fully tested it yet, but my preliminary testing seems to suggest it is currently working.

Adds versions of the buttplugs that have tails.


prndev:
I like the idea of buttplugs with tails, but this adds a lot of overhead in checks and redundancy in code. I suggest we should split the buttplug property into multiple properties buttplug ("none, plug, large plug, huge plug") for the plug size and buttplugType (unset, "tailed") for the type.

freecitiesbandit:
Agreed, though I suspect the proposed fix might be out of my skill range.

PR #787, opened 2017-07-04
Fixed degrading name bugs

kopareigns:
If a player chose to have degrading names applied automatically to new slaves, and did not have a sufficiently progressed revivalism future society, buying slaves from the household liquidator would throw an error in the widget NationalityToName which originated in the call to the degradingName.tw passage. The problem was that $activeSlave was unset and degradingName.tw only changes $activeSlave. I just changed it to make sure $activeSlave was still set, and not pointing to the first slave.

I've opened this in a new pull request because I'm not sure this is the ideal solution. A more robust fix might be to make degradingName its own widget instead. If I have time I'll try to get a working version of that ready.


kopareigns:
I'm closing this as I feel confident I have a working widget for this. I expect to have a new pull request for that soon.

PR #788, opened 2017-07-04
Lifestage between 25 and 35

freecitiesbandit:
It always bothered me that all slaves between 25 and 35 were usually just titled 'slave'. This adds a new stage for slaves aged between 30 and 35, 'mature'. I just wanted to post this on here to get a read on what people thought about this title. I don't think it's perfect, but I've had trouble thinking of other titles. Let me know if you feel it is unnecessary as well.

Other ideas: slavelady, slavewoman, cougar.


tc-k:
Cougar and mature as used in porn don't match with the 25-35 age category. Those two terms overlaps with MILF or even GILF. Urban Dictionary suggests the term wildcat, but honestly that term is not popular.

The normal term is just women, but in the context of this particular fetish I think this is one case where the plain "slave" is the right term. If you don't think that's descriptive enough, slavewoman can be used. Slavelady is weird though. Sounds like some paternalist neologism.

freecitiesbandit:
I agree largely, the various possible terms seem rather imprecise. I think slavewoman is the most neutral sounding, but also doesn't quite convey that 30-35 age grouping. Do you think mature fits at all, or does slavewoman fit the bill better?

Also, upon rereading your comments, this new title only covers the age group of 30-35. 25-30 will still be described as slave, cow, amazon etc.

tc-k:
For me when I see the word mature in a porn category, I expect women (and I suppose men) who is between 40-60. That keyword in a porn site is likely to get you some grannies hit.

I am actually not sure what's a good term for women 30-35 years old. I guess MILF, but that's already in use and really MILF's age range is quite wide too. If you must differentiate that age category from MILF, I guess slavewoman is the least bad.

Hmmm, maybe ask the FC community at reddit?

freecitiesbandit:
Good idea, I just posted a thread.

freecitiesdev:
The reason the titles are in 3 categories (18-24, 25-34, 35+) is that these are mechanically significant in various ways. I'm not in love with the suggested titles, and I'd have to be in order to add title categories that have no mechanical impact.

freecitiesbandit:
Makes sense, I haven't been getting any great ideas from reddit either so I think we can safely close this pull request.

PR #790, opened 2017-07-05
DegradingName widget

kopareigns:
Incoporated degradingName.tw as a widget in miscWidgets.tw and deleted degradingName.tw. This removes a few possible errors and should provide a more robust system in the future for renaming slaves this way.

I've tested in all three different ways this widget can be called and I haven't run into any errors or obvious problems.


freecitiesdev:
This was on the to-do list but you beat me to it. The same goes for a lot of obvious widgetization of old code, so if you're up for it, go nuts.

kopareigns:
I'll keep that in mind next time I wade through the code. Are there any specific areas/passages you'd like me to prioritize? Is there a general type of code that should be widgetized? E.g. code that appears often/rarely, description code, slave generation code, etc.

I only ask because I'm a novice at coding in general (especially at javascript/twee) and don't really know how to identify areas where using a widget is preferred. The degrading name passage only caught my eye because it was actually throwing errors.

fcpmanon:
Any identical code that is copy-pasted in multiple locations is a good candidate to replace with a widget.  Duplicate code slows down development as multiple places in the code must be kept consistent to avoid errors.

Beware of variables used in widgets or in passages that call widgets or in passages that are included by widgets.  If you change a variable in a widget, the passage that called the widget will have the modified value for that variable afterwards.  This can cause unintended effects when widgets are called inside for loops and something in the widget (or in another widget or passage included/called by the widget) changes a variable that the calling passage's for loop is using for other purposes (like _i).

The correct way to pass data to widgets is to include the data in the widget call, like this:  `<<widget "WidgetName" param1 param2 param3 ...>>` Then in the widget you can access the param values as $args[0], $args[1], $args[2], ... (notice first param is number zero).

kopareigns:
Thanks for the advice. I have one more question if you don't mind. Do you know if there are there any advantages to converting certain passages entirely to a widget like I did for the degrading name passage here? For example the "Remove activeSlave" passage is called fairly frequently and could be widgetized, but it wouldn't really reduce the amount of duplicate code, just change it from an included passage to a widget.

freecitiesdev:
>Do you know if there are there any advantages to converting certain passages entirely to a widget like I did for the degrading name passage here?
I don't have the expertise to answer this, but I have asked several other people to look into the matter. Generally the answer has been "not really." Furthermore, I'd recommend leaving things like "Remove activeSlave" alone unless you're quite confident in your coding and testing skills with Twine. Updating the $slaves array inside widgets can go sideways in strange ways.

kopareigns:
Another thing I'll keep in mind. Again, thanks for the advice.

PR #803, opened 2017-07-07
Added display of target race for raiding event.

freecitiesbandit:
Event now lists target's race.


freecitiesdev:
Good work, which I rarely bother with. Some players apparently care deeply about this, but it's often a pain in the ass.

PR #804, opened 2017-07-07
A truckload of updated lisps

kopareigns:
Replaced a bunch of "checkForLisp" if/else statements with the corresponding widgets. Something I didn't realize immediately is the <<Master>> widget auto-uses $activeSlave if there are no arguments (and resets all other pronunications to be for $activeSlave too). So if you want another slave to use it properly, say $subSlave, you always need to use it as <<Master $subSlave>>.


freecitiesdev:
That's concerning. <<Enunciate>> is supposed to set all pronunciations until the next <<Enunciate>>. It's rerolling $s, $ss, etc. for each <<s>>, <<ss>>  called?

kopareigns:
Just the `<<Master>>` one resets it. If `<<Enunciate $slave>>` is called, it'll properly set all the `$__Enunciate` variables. Each of the small `<<s>>` or `<<S>>` type widgets then just prints the variables.

However, `<<Master>>` is built so that it may or may not accept an argument. If it doesn't have an argument, it runs `<<Enunciate $activeSlave>>` before printing anything. So to get `$subSlave` (or anything thats not `$activeSlave`) you have to run `<<Master $subSlave>>`

I didn't realize the markdown formatting accidentally ate some of my text above until now, likely making it more confusing to read. Sorry if that was confusing.

PR #806, opened 2017-07-07
Added basic intro content for cumslut

lowercase-donkey:
Always seemed weird cumsluts didn't get an intro option, so here's one for inflation fans.

I based it on spanking/masochist, and am looking for suggestions on how to improve it.  In particular, should I tie it even more to the dairy somehow?


freecitiesdev:
It looks good. If anything, you could split it into two versions, a vanilla one, and one that's only available if the Dairy is installed and/or penthouse milkers are available.

lowercase-donkey:
Done.  #809

PR #809, opened 2017-07-08
Industrial dairy inflation

lowercase-donkey:
Inflation is always a weird fetish when it comes to "realism," and I don't want this to be too immersion breaking.  I've tried to indulge fantasy as well as nodding to the slave's safety with this magical 'gel,' but that might be too much Handwavium for Free Cities if we "stretch" things too far.

I've also tried to leave the master a bit free in his fancy as he watches, leaving players to fill in the image as they please.

Especially if this is your fet, please feel free to suggest improvements or make your own.

I've checked it with a few slaves now, caught some bugs, seems good.

That means I'll commit and find something wrong immediately.  D:

### DIFF
--- a/src/uncategorized/newSlaveIntro.tw
+++ b/src/uncategorized/newSlaveIntro.tw
@@ -1398,60 +1398,60 @@ The legalities completed, ''__@@.pink;$activeSlave.slaveName@@__'' <<if ($active
 			<</replace>>
 		<</link>>
 	<</if>>
+
+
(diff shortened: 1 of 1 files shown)


lowercase-donkey:
Since I'm starting to write a little bit, I guess I'll add here that I don't mind being credited under this name, if it ever gets to that level.

freecitiesdev:
Feel free to edit yourself into the credits.

PR #813, opened 2017-07-08
Lisping Round 2: Electric Boogaloo

kopareigns:
A bunch more updates for lisping. I also modified the `<<Master>>` widget so that it only defaults to $activeSlave if $titleEnunciate is undefined. As best I can figure, this should keep it compatible with all current uses of the widget.


kopareigns:
Update: Also modified the `<<say>>` widget to allow for options other than "say" if the slave has no lisp. If the slave has no lisp, typing `<<say shout>>` will print "shout" while typing `<<say>>` will still print "say". If the slave has a lisp it will always print "lisp".

freecitiesdev:
Excellent idea.

PR #815, opened 2017-07-09
Fixes

kopareigns:
Updated some old values and tweaked surname generation for related slaves.

Edit: and tweaked the wording a bit in the new interslave begging event so that there weren't any weird mangled sentences.


kopareigns:
Also clamped the minority shares of an arcology to be between 0 and 98 in a certain case. This *should* fix the issue of the player and their rival being able to have a combined total greater than 100% of arcology shares. 

If I parsed the calculations correctly, it was previously possible for the minority to have negative shares, which in turn allowed the player to purchase two more shares no matter the share belonging to your rival. This was only possible with the rival arcology.

PR #822, opened 2017-07-10
Restructuring of belly description widget.

pregAnon:
It now focuses on the clothing rather than the size of the slave's belly, much like the other desciption widgets focus on the clthing first.  Makes it much easier to add new outfits, since all the descriptions for it are now in the same place.  Also will allow for text consolidation.  Lastly, cleaned up slightly off .weight checks.

Not an absolutely necessary merge, but makes it easier to work with I think.  The old set up was easier to work when initially coding it, but doesn't provide any benefit to, if not complicates, new clothing addition.

### DIFF
--- a/src/utility/descriptionWidgets.tw
+++ b/src/utility/descriptionWidgets.tw
@@ -6943,710 +6943,770 @@ $pronounCap has
 <<if $activeSlave.preg > 30>>
 	$pronounCap is enormously pregnant,
 	<<if $activeSlave.height >= 185>>
-	but $possessive tall frame bears $possessive massive, drum-taut belly well.
+		but $possessive tall frame bears $possessive massive, drum-taut belly well.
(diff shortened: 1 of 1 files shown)


freecitiesbandit:
Excellent idea and implementation. I've been looking into working new clothes into FC but the sheer volume of code changes required to implement one item has been putting me off.

pregAnon:
I'm pretty sure I'll need to rework walkPast's belly description as well.  I'll handle that when it gets in my way, shouldn't be too long.

lowercase-donkey:
Yeah, it's crazy how hard it is to add clothes.  And I wanted to add slots like gloves! This should help, thanks!

freecitiesdev:
Clothes and walkpast are some of the oldest surviving artifacts of FC 0.1. By which I mean to say, Godspeed.

PR #824, opened 2017-07-10
Major bodyMod changes to layout and code

lowercase-donkey:
Major work on bodyMod to condense the page and (I hope) make it easier to adjust values in code.

So far:
Tattoos
Piercings
Branding

Tattoo work is great if I say so myself, but piercing I have mixed feelings on.  It's very clumsy trying to make the descriptions into sentences when any part or all of them might be missing.

I made #825 as a version that has the new "choice" style of options, but mostly keeps the simple list of piercings.  (edit: it's quite out of date now)

Lastly, a hybrid approach could look quite good.  Simple piercings look the best compressed, but the long descriptions would work better spread out in a list.

PR #825, opened 2017-07-11
Compare version: "List" version of piercing descriptions

lowercase-donkey:
This is the old style list, but with "unpierced" stripped out.  It's not polished.

It won't really save space if she is pierced, but still saves quite a bit if she isn't.  

Don't merge this and #824 , please, just pick one.  Whichever is better.

Edit: #824 is now slightly ahead with bugfixes.


lowercase-donkey:
Notably improvements to corset would need to be ported over.  More a note for me, if anything.

PR #827, opened 2017-07-11
Re added Asian art .lipsTat

pregAnon:
There exists a discription for it, so it would be counterproductive to remove it.
Conversely, there is no .dickTat for "Asian art".


lowercase-donkey:
I dropped it in #818 per @freecitiesdev comment on #778.  I'm fine with reversing that commit as long as it's available in BodyMod too, which it now is.

You might want to hold off on this until after my bodyMod changes get reviewed, I changed, well, most of the page.

pregAnon:
Go for it.  Though I'd also say integrate this request into yours to prevent the inevitable conflict.  Should be simple enough, plus will give me less of a headache trying to place it afterwards.

PR #828, opened 2017-07-12
Re added Asian art .lipTat

pregAnon:
It has become abundantly clear that I have no idea how to resync my local version with github short of ripping out the offending file and replacing it.  It works though, even if it makes needless commits.

### DIFF
--- a/src/uncategorized/bodyModRulesAssistantSettings.tw
+++ b/src/uncategorized/bodyModRulesAssistantSettings.tw
@@ -1080,7 +1080,7 @@ __Rule $r tattoos__
 <</link>>
 |
 <<link "Asian art">>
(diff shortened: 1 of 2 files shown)


freecitiesdev:
Don't worry about it. As long as you're not submitting needless PRs, you're ahead of the curve.

Also, are you PregModder? If so, shoot me an email. I dropped you one a couple of weeks back and didn't get a reply.

pregAnon:
I act as a go-between of sorts.  So I'll pass along a "look at your e-mail for once" message.

freecitiesdev:
It's entirely possible he's not at home on purpose, in which case, fine. But do let him know I'd like to chat.

PR #833, opened 2017-07-12
Some small fixes for the new event

kopareigns:
Also, is it intended that there are no changes to the slave in the egyptian revivalist case? Currently it's the only case where the slave is entirely unchanged.


pregAnon:
Look at "broadening trainer" too.  "D cannibal" is nested in it and I don't see how that will even work.

freecitiesdev:
No, it isn't. I'll take a look.

PR #835, opened 2017-07-12
Create widgets for piercings and tattoos

lowercase-donkey:
All that space I saved...  

Oh well, this is better.  By moving all tattoos and piercings into widgets, the content can be viewed in contexts other than fleshy viewing.

The obvious place to apply this is bodyMod, which is now longer again, but we actually get accurate descriptions here.  I think it's good flavor.

Edit: also made widgets for nails and makeup while I was here.  It was supposed to be the same text in salon and longDescription, but it had started to diverge a bit.


src/uncategorized/bodyModification.tw
```diff
@@ -21,7 +21,8 @@ $activeSlave.tonguePiercing +
 $activeSlave.nipplesPiercing + 
 $activeSlave.areolaePiercing + 
 $activeSlave.corsetPiercing + 
-$activeSlave.navelPiercing + 
+$activeSlave.navelPiercing +
+$activeSlave.corsetPiercing +
```
racerand: CorsetPiercing seems to be counted twice

racerand:
In line 25 of bodymodification corsetPiercing seems to be counted twice

lowercase-donkey:
Thanks, I'm moving over to a new fork and made changes there.

freecitiesbandit:
I can't for the life of me figure out how to correctly implement new tattoos. I added new descriptions to descriptionwidgetstattoos and then added the ability to add them in bodymodification.tw, but when I try to apply them they just create tattoos with blank/empty descriptions. Is there somewhere else I need to be adding stuff?

PR #848, opened 2017-07-14
Persistent Summary

DrNoOne0:
A rework of how the game handles the individual slave summaries.  Instead of creating each slaves summary from scratch each time it is displayed this makes it use a stored string instead, generating a new string when needed.

Using a 190 dairy cow save for timing, the first dairy load averaged ~15% slower (16.5 -> 19.5), but subsequent loads were ~40% faster (15 -> 9).  More speed can probably be found by reducing dependence upon the Slave Summary passage.

Situations that makes the game decide it needs a new summary for a slave:
- Any newly created slave
- All slaves at the end of the week (during the Next Week passage, after events)
- A slave's inspection page (covers personally changing slave's rules, drugs, or applying surgery)
- Toy Box options (bottom of main page)
- Manually Re-applying Rules Assistant (from main page or RA options page)

While I have tested this, I haven't been able to try all summary combinations.  It should generate functionally identical summaries to the current system across all option settings and slave values.

### DIFF
--- a/src/gui/Encyclopedia/encyclopedia.tw
+++ b/src/gui/Encyclopedia/encyclopedia.tw
@@ -942,7 +942,7 @@ __I do not give credit without explicit permission to do so.__ If you have contr
 ''Rodziel'' contributed the cybernetics mod.
 ''prndev'' wrote the Free Range Dairy Assignment scene.
 ''freecitiesbandit'' wrote a number of random events.
-''DrNoOne'' wrote the bulk slave purchase code.
+''DrNoOne'' wrote the bulk slave purchase and persistent summary code.
 ''Mauve'' provided vector art for chastity belts and limp dicks.
 
 ''Many other anonymous contributors'' helped fix bugs via GitHub. They will be credited by name upon request.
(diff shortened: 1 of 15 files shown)


kopareigns:
Sounds pretty neat. Only thing I can see without spending hours digging through all the changes is that you don't have anything for the new tailed buttplugs.

pregAnon:
How will this work out for players that don't frequently enter and exit facility screens, etc?  Will they just see the 15% loss and no gain?

DrNoOne0:
@kopareigns Added the tailed buttplugs, thanks.  I also looked through the change history for slaveSummary.tw for any recent changes, before the buttplugs was surname, and I have that, so hopefully it is now good.

@pregAnon It is possible.  To get just the 15% loss you have to load the facility once and only once before the slaves need updating.  If you do anything that causes a reload of the page (like assigning a slave to the facility) you'll make the time back.

freecitiesdev:
After some thought, I'm merging this for now. I can't really come up with a reasonable way to play the game and have this be a noticeable loss, and there are a lot of ways to play that are going to see a huge gain.

My one major question is that this is speeding up the main menu, isn't it? There are a lot of operations that involve revisiting Main over and over, and as long as there's no RA changes, this is speeding that up, correct? I think it is but I don't have a 100 penthouse slave save handy to test it thoroughly.

DrNoOne0:
It should be speeding up any page that displays slave summaries, the main menu included.  If there are no slave summaries shown on a page, there should be neither a performance loss or performance gain.

PR #849, opened 2017-07-14
Headgirl reprioritization

kopareigns:
Motivated by @buttwhy and since I was kind of thinking of doing this too, this is a rework of the way the headgirl decides which slaves to train first. She will now make sure every slave is healthy before moving on to removing paraphilias, before moving on to fixing other flaws, before moving on to softening, etc. Assuming she's allowed to train those at least. I've tested a few combinations of trainings and it seems to work as intended.

Of note, I avoided using a bunch of
```
<<if $HGEnergy == 0>>
    <<break>>
<</if>>
```
statements by including an extra conditional inside the for loop statement. I'm not sure when else this would be useful, but I guess this is confirmation this technique works.

If preferred, these calculations could also be moved to the beginning of Penthouse Report or put behind a game setting.

### DIFF
--- a/src/uncategorized/fullReport.tw
+++ b/src/uncategorized/fullReport.tw
@@ -91,68 +91,12 @@
 	<br>
 <</if>>
 
-<<if ($HGEnergy > 0)>>
-<<if $slaves[$i].fuckdoll == 0>>
-<<if ($slaves[$i].fetish != "mindbroken")>>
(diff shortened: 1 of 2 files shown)


freecitiesdev:
This sounds like an improvement but I'll have to playtest it quite a bit. With the more robust dialog widgets, I've been wanting to let HGs tell you what they're doing in their own voices, and this would be a good opportunity for me to start working on that,

PR #851, opened 2017-07-15
Made compiler work while running as admin

IWantToBeAnonymous:
If you run as admin on the batch files, the working directory is set to C:/Windows/System32. The previous echo command would also include extra text.


lowercase-donkey:
Running as admin still doesn't work for me.  If you edit "dir" in after the cd, you can see it's still scanning system32.  I can give more info if needed, but it's puzzling.

PR #855, opened 2017-07-15
Cheatmode/debug update

kopareigns:
Updated RE recruit and RE FS acquisition to include functionality for cheatmode. Also changed the huge if/else statement to be a switch in RE recruit.

### DIFF
--- a/src/uncategorized/reFSAcquisition.tw
+++ b/src/uncategorized/reFSAcquisition.tw
@@ -1,12 +1,27 @@
 :: RE FS acquisition
 
+<<if Array.isArray($FSAcquisitionEvents)>>
+  <<if $cheatMode == 1>><<nobr>>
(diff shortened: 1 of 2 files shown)


lowercase-donkey:
Programming question, what's the advantage of switch over if/else?  I'd like to know when to use each.

kopareigns:
I'm far from an expert in programming, but the simple googling I did earlier seems to imply that switch can be faster if you have a significant number of possibilities, although this number probably depends on the language and implementation. In this case since we're not iterating, the difference is likely negligible.

However, the recommendation was to use switch/cases if you were in a scenario in which you definitely have certain cases. E.g. the possibilities for $recruit in this case. So using a switch seemed the proper programming etiquette in this file.

freecitiesdev:
My understanding is that the efficiency benefits kick in around 5-6 cases. There are lots of places in FC with more than that: looking at you, societies.

PR #856, opened 2017-07-15
Add links to rivals and relations, color them appropriately.

lowercase-donkey:
I hope it's ok to add this to the stylesheet, it's the only way I can get the color of the link to respect the color of the passage.

This will be useful elsewhere I'm sure, but it's also technically possible that someone else was assuming links would always be blue, and encased one in @@.  If that's the case, it's just a matter of correctly escaping the link itself from the palette choice.


freecitiesdev:
I'm merging this for now but it's liable to be revisited. I'd like to have an open discussion about link colors at some point. There are definite arguments to keep a single link color, and arguments for a more open system.

PR #858, opened 2017-07-16
More fixes

kopareigns:
A couple missing "set"s in the new summary update, walkpast no longer tries to mess with an undefined slave at the start, and a double counting I noticed in the cybernetics mod.


DrNoOne0:
I thought I had fixed those sets.  I guess not.  Thanks!

kopareigns:
Also removed some widget calls from the new wardrobe that depended on being in the slaveInteract passage.

PR #861, opened 2017-07-16
Simplify wiki navigation

lowercase-donkey:
The glossary list was uncomfortably long.

There were three ways to nav before which got confusing as well: Glossary, Permanent link section at bottom, Dynamic link section at bottom.

This commit leaves only a permanent section on left for umbrella articles (where glossary was; it was permanent anyway) then links to more narrow articles in the dynamic section bottom of each article.

I added a few of the "umbrella" terms such as "terrain" as placeholders and the dynamic links at bottom.

TL;DR, more friendly to new masters and old alike.

Minor fix: missing "" in encylopedia for a while too. What do these do?


lowercase-donkey:
What the heck happened to "collection facility?" There's references to it everywhere.  I'm assuming cockmilking is just included in milking now?

There's very little that links to it, but references are still there.  Going to remove in a commit.

kopareigns:
Used to be a separate facility. Was merged with the dairy in 0.6.8.

PR #862, opened 2017-07-16
Old event pruning/streamlining/randomization

kopareigns:
Just went through a bunch of event files, removed a bunch of extraneous code, and added a couple tiny fixes.


kopareigns:
Update: Added randomization to the relevant RETS events that were not yet randomized. Also simplified some of the for/if stuff in the opening of RETS where the event slave was guaranteed to not be in the subID list. Fixed a typo, and modified some checks so that the RETS events only have a chance of firing if there are two different eligible slaves.

freecitiesdev:
Excellent work. You beat me to a few of these.

PR #863, opened 2017-07-16
New tattoos

freecitiesbandit:
Three new tattoos and I also found remnants of a sacrilegious tattoo, so I fleshed that one out too.

New tattoos: 
Paternalist (to contrast the degradation tattoos) 
Possessive (tattoos focused on your ownership)
Sacrament (tattoos that might suit a chattel religionist arcology

Expanded tattoos:
Sacrilege (tattoos that have a more heretical bent to them)

Thank you @lowercase-donkey  for making this so much easier to do.

### DIFF
--- a/src/uncategorized/bodyModRulesAssistantSettings.tw
+++ b/src/uncategorized/bodyModRulesAssistantSettings.tw
@@ -1133,6 +1133,114 @@ __Rule $r tattoos__
 <<set $currentRule.legsTat = "scenes">>
 <<RAChangeLegTattoos>>
 <</link>>
+|
+<<link "Sacrilege">>
+<<set $currentRule.lipsTat = "sacrilege">>
+<<RAChangeFaceTattoos>>
+<<set $currentRule.shouldersTat = "sacrilege">>
+<<RAChangeShoulderTattoos>>
+<<set $currentRule.boobsTat = "sacrilege">>
+<<RAChangeChestTattoos>>
+<<set $currentRule.armsTat = "sacrilege">>
+<<RAChangeArmTattoos>>
+<<set $currentRule.backTat = "sacrilege">>
+<<RAChangeBackTattoos>>
+<<set $currentRule.stampTat = "sacrilege">>
+<<RAChangeStampTattoos>>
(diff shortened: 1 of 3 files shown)


lowercase-donkey:
Great work here.  Maybe I am just missing it, but maybe add a society check for tats that are tired to a society.  Take a look at the way clothes work.  Typically you need to Roman Revivalist to get Togas, but you can buy them through wardrobe IIRC.

If you get really ambitious, maybe allow "tattoo designs" to be purchased in penthouse->wardrobe as well?

But (thinking out loud now) that sort of gets limited by the "wardrobe" nature of wardrobe.  Maybe wardrobe should be made more universal. @pregAnon, wardrobe was your baby, wasn't it?  What do you think?  Rebrand it as some sort of Goodsforyourgals.com?  Where did you envision these clothes coming from?

freecitiesbandit:
Degrading tattoos aren't set to be associated with degradationism are they? Though I suspect the degrading tattoos are a little less FS themed as they are just generally degrading. 

Though in the interest of not limiting content for the PC, you'd have to be able to 'buy' FS associated tattoos, which seems a little arbitrary.

pregAnon:
Wardrobe is more for physical objects.  New tattoos would likely be some sort of downloadable design pack for the body mod studio.  As for contact designs, I have no ideas, though alternate cybernetic eye colors would likely fit into the research lab, since you can't get it without cybermod active.

lowercase-donkey:
Ok, I agree with you then.  I can see how wardrobe would naturally include even dildos and chastity belts.  The only physical thing slaves get outside of Wardrobe would be piercings, but I really don't think we're getting fancy in piercing styles yet.  It's pretty much "yes" or "very" so far.

No need to change.

freecitiesdev:
I'm deferring final disposition of stuff like what goes in the wardrobe until it's seen a public update, which I'm trying desperately to get out ASAP.

lowercase-donkey:
@freecitiesdev can you expand on that?  Wardrobe is getting a public update?  Or FC doing major release?  Or what?

PR #864, opened 2017-07-16
New Eyes

freecitiesbandit:
Added a few new types of artificial eye shapes:
Wide-eyed
Almond-shaped
Bright
Teary
Vacant

These are a little more mundane than the existing eye shapes, but the hope is that they allow for more descriptive eye descriptions.

Also took the opportunity for vanity and updated my encyclopedia entry.

@lowercase-donkey  Looks like the RA Salon doesn't work with your new eye color/shape system.


lowercase-donkey:
> @lowercase-donkey Looks like the RA Salon doesn't work with your new eye color/shape system.

Shiiit.  I'll get on it.

freecitiesbandit:
Thanks for the quick work. I think that should make it work though?

PR #867, opened 2017-07-18
Menstrual Cycle

IWantToBeAnonymous:
Added menstrual cycle to each slave as described in the google doc. Probably need to flesh it out some more, but it's a start.


pregAnon:
I was planning something like this and built a contaning widget to handle it.  I'll look into getting it in soon to greatly simplify impreg checks.

IWantToBeAnonymous:
You're right, it does need more work. 

About the fVagina scene, unless I missed something, doesn't the check to `$activeSlave.preg === 0` cover it if they're on contraceptives?

pregAnon:
Better to use canGetPregnant($activeSlave).

IWantToBeAnonymous:
Ah thank you. I need to get to know the codebase a bit better it seems.

kopareigns:
Actually I think you got me on that one. It should catch contraceptives like that. That was my turn to not be 100% familiar with the codebase. I hope I didn't come off too negative previously. It's always good to have more people willing to contribute.

IWantToBeAnonymous:
I don't think you did. I understand not wanting to merge something that is unfinished/not completely fleshed it.

pregAnon:
Biggest concern I have is I'm planning an extensive rework of the entirety of the conception/pregnancy system.  I think the best course would be to get you hooked up with the aforementioned conception widget and tweak things afterwards.  Would be far easier to call the widget than that whole conception check every time.  Let me go get that set up and ready for your use, though FCDev will likely get it in tomorrow evening.

pregAnon:
Said widget is now viewable in my pull request.

IWantToBeAnonymous:
Does the PC have a specific ID I could use? Or should I just supply -1 instead of 0?

pregAnon:
$PC's ID is effectively -1, but in this case it is better to just use -1.

pregAnon:
Another suggestion.  Instead of freshly generated XX slaves always being on their preriod, it would be better to random(0,4) it and give a generic .pregChance to them based off of it.  Will give a bit of variety out of the gate, you know?  Also, as it is now, it will likely break slave intros like "impregnate her".

IWantToBeAnonymous:
I'd thought about setting it to a random value. Yeah, there is quite a bit I need to do before this is ready. As well, the slaves you receive from taking over the arcology don't have the variables defined either.

pregAnon:
If they get built off of generateXXSlave, it should be set.

pregAnon:
Unless you are referring to hero slaves, in which case, a simple setter in acquisition can take care of that well enough to handle the first week.

pregAnon:
Alternatively, removing the check from generateXXSlave and placing it in the "AddSlave" widget would work well too, as it will catch and randomize every added slave's menstrual cycle.

pregAnon:
Looks good.  The last things that I had noted down are anorexia and steriods interfere with menstrual cycles.  With that, my notes are tapped out nearly completely.

IWantToBeAnonymous:
Yeah, seems to make periods irregular. I could have it check for either of those and not increment the cycle value based off of some random number. Not really sure though.

freecitiesdev:
See #876.

freecitiesdev:
Two minor requests, neither of which is preventing a merge.

1. I'm not a huge fan of calling a minor addition like this "Menstruation Mod" in-game. FWIW, both major mods that are still called mods in-game are, in my opinion, in need of either being heavily revised or removed. Not good company.

2. Contraceptives ought to stop menstrual cycles. They're implied to be hormone-based, and that's more or less how those really work for many women. This will also give players who like realism but dislike menstrual cycles a way to realistically bend their slaves to their will. (The whole point.)

PR #845, opened 2017-07-13
remove parentheses in case statements

goldifrocks:
some cases were merged, hopefully i missed nothing.

PR #870, opened 2017-07-18
Conflict Resolution

kopareigns:
Just a replication of some (but not all) fixes in #845. Should allow for easier merging.


freecitiesdev:
Closed #845 due to overlap.

PR #879, opened 2017-07-19
Event slave randomization

kopareigns:
Added more random slave selection to relevant events. Looking at doing a few more, but waiting until I feel I have a better solution for those.


kopareigns:
Question: Why are some slave objects set to 0 and some set to null in nextWeek.tw? Is there a functional or moral difference between what gets zeroed and what gets nulled?

kopareigns:
There are the relationship events. Of note, I filtered a smaller _relatedSlaves array from $slaves first and then used filter/find with that array. Since this array should be smaller than $slaves, this should be more efficient than using the $slaves array unless it's nearly the size of $slaves.

What other events could be randomized like this? The only one that comes to mind right now is P mercenary romeo, and I'm still undecided on how to approach that.

freecitiesdev:
>Question: Why are some slave objects set to 0 and some set to null in nextWeek.tw? Is there a functional or moral difference between what gets zeroed and what gets nulled?

There's no functional difference that I'm aware of for things like slave objects. Either is going to reduce the size of the autosave, which is the point. The only things that need to be 0 are counters: variables that might be added to without being set yet next week.

PR #893, opened 2017-07-21
slaveSurname Tweak

kopareigns:
Another tweak to another commit I made a little while back.

I feel I may have misunderstood the original intent of how slaveSurname was supposed to be assigned. This tweak will make it so the default setting is for slaves to have a surname regardless of whether $useFSNames is toggled to 0 or 1. If it was intended for the default to be that slaves automatically get no slaveSurname, or if was supposed to depend on $useFSNames for some reason, I can make another small tweak before this gets merged.

Perhaps we should make an independent setting for whether or not slaves get stripped of their surname upon enslavement? Edit: Or perhaps a policy?


freecitiesdev:
>Perhaps we should make an independent setting for whether or not slaves get stripped of their surname upon enslavement? Edit: Or perhaps a policy?

I think it's better as a rule than as a policy.

PR #896, opened 2017-07-21
Minor: missing line breaks in clinic

lowercase-donkey:
Not sure how I missed that


lowercase-donkey:
Just check spa as well as the other rooms and they look fine.  Guess I had spa on the brain.

PR #899, opened 2017-07-21
random assorted fixes

goldifrocks:


DrNoOne0:
> _slaveName is not used in widget
Yeah, that should be removed.

The reason why it was there was that originally I thought was going to have the slave's name in the summary.  Then I realized not all summary uses had the name link lead to the Slave Inspect passage and scrapped the plan.  I just missed removing that part of it.

PR #901, opened 2017-07-22
Quick Summary Options page

DrNoOne0:
Uses magic to make the Summary Options page extremely fast for changing summary options.  By magic, it is actually a combination of widgets, spans, and links with replace statements.

It also rearranges the order of options on the page to match the order the options they control show up in the summary.


lowercase-donkey:
God I love widgets.

PR #903, opened 2017-07-24
Extended Family Mode + extras

pregAnon:
Extended family mode allows for multiple family members, go read the documentation for more details.

Also included are an unreasonable amount of bugfixes, oversight catches, code clean up and other little things, new oversized sex toys, an overhaul of the buttplug system to prevent future code bloat, and a bunch of new utilities.
Said widgets are "PlayerName", "SlaveFullName", "PoliteRudeTitle" and "WrittenMaster".

To note: walkPast's call them both over doesn't work with extended family mode, starting girls lacks a proper implemenation for utilizing said mode, and slaveSummary's new style was not very compatible. It will need a bit of prettying up by someone who understands it better. But for now it is functional.


lowercase-donkey:
Really impressive work!  So what happens with tail plugs now?  Are they the long plugs the code refers to, or are they handled separately?

PR #892, opened 2017-07-21
Updated Bovine Tattoo descriptions

freecitiesbandit:
@lowercase-donkey I just went ahead and created a new pull request for the changes. Thanks for your previous work and fixing the issues beforehand.

PR #883, opened 2017-07-20
Fix amputee bug in bodymod; restore missing bovine tats to bodyMod, fix many missing bodymod checks in longSlaveDesc

lowercase-donkey:
From [reddit](https://www.reddit.com/r/freecitiesgame/comments/6oc7sk/not_sure_if_bug_or_purposeful_omission/).

Edit: while in here, I found I failed to add bovine tats when I overhauled bodymod, I think.  Whoops.

PR #913, opened 2017-07-29
Better Bovine Tat descriptions, removed special/universal bovine tat …

lowercase-donkey:
…descriptions (edit: and now Create brandDescription widget, bug fix)

Bovine tats now have enough descriptions to stand on their own, and no longer need weird kludges.  In particular under the older code, a slave with a "shoulder" bovine tat would have her arms and back described with bovine tats, as well as whatever tats they actually had. Unfortunately, slaves could look like they had "two" tats on their arm this way.  I recently tried a few fixes, but thanks to the creative descriptions of @freecitiesbandit in #892, this is no longer needed.  This branch replaces #892 completely.

This commit also allows bovine tats to be applied piecemeal just like other tats and still be described accurately, so I partically reversed my commit in #883.

TL;DR Bovine tats are now as boring as tribal tats.

### DIFF
--- a/src/uncategorized/bodyModification.tw
+++ b/src/uncategorized/bodyModification.tw
@@ -1289,7 +1289,8 @@ Choose a tattoo style:
 
 Branding: 
 <<if $activeSlave.brand != 0>>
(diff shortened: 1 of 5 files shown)


lowercase-donkey:
Sorry, I meant to make the second commit a new branch.  Should be simple though.

PR #914, opened 2017-08-01
Wiki work

lowercase-donkey:
Sort code in wiki to be grouped and in the same order as it appears in the game, making it far easier to find mistakes and add future content.

Add stubs for Clit, Waist, Hips, and Areolae.  Clit in particular is going to need some TLC, since it interacts with many systems, like Dicks do.

Merge a few "lore" style entries that were lost in the shuffle and likely have been for some time.  These focused on roles, so there was a more technical head girl page, and then a page with a lore blurb about HG.  Repeat for madam, etc.  Lore now appears with appropriate descriptions of game mechanics.

Change quoted/lore text to be in italics, so it's more clear what's flavor text and what's game stuff.  Think Civilopedia.

Fix many minor mistakes.  Still haven't settled on caps for linking roles and places, but caught a few more caps problems.


freecitiesdev:
Outstanding work.

PR #915, opened 2017-08-01
Starting slave support for Extended Family Mode, Peacekeepers Lategame

freecitiesdev:


Weresmilodon:
Might have put this in the wrong place?

I noticed that in 'Provide immediate assistance' and 'Deliver generous aid' one sum is written into the event text, and another is used as actual cost. The warning is the same as the cost in both cases. Not sure which one you intended to use.

Both are in pPeacekeepersIndependence.tw of this commit.

PR #918, opened 2017-08-02
Switch to the new `nobr` global; remove nobr tags where they appear

lowercase-donkey:
This is one less thing to worry about now.  Most code was using `nobr` anyway, once the wiki was updated in #917.  Should make life easier for newbs and old timers.

Obviously this touches a LOT of files, but changes to each are minimal.  There were surprisingly few cases where I had to add `<br>` tags, but it's likely I missed some places that we'll have to catch later.

Minor: in the process of removing the backslashes, I found a lot of stray ones (one item in a series of ten might have a backslash.)  I believe they were inert, since it looks like `nobr` ignores them anyway.

Sorry for the excessive commits, I'm not sure how to `git grep` as though I had already committed.


ObstacleCorpse:
I'd be more than a little hesitant to accept this pull request. Making the nobr tag global will have detrimental consequences to any mods/forks with custom passages, as they will all have to be adapted to this new format, which provides only a minor convenience in exchange.

Additionally, TweeGo (the compiler) is about to be updated after over a year of development. I would at least wait until the changelog for that is published (and the compiler is tested) before making such a drastic change.

lowercase-donkey:
Well, I agree it is a large change in so far as it touches a lot of pages.  At worst, though, for mods it's a matter of search/replace a document for double link breaks and replacing them with `<br>.`  Having just done it for the entire game with notepad++, the work is trivial.  You'd be surprised.  Notepad can even do it for files across whole projects, so it's just firing one macro, basically.  

There is a serious advantage in having something consistent for forks and mods to work with as well.

I'm not sure what kind of changes are anticipated in the compiler that would impact this.  Are you thinking there will be a per document escape for the global or something?  Are you thinking `nobr` will be depreciated like `silently` was?

freecitiesdev:
>Additionally, TweeGo (the compiler) is about to be updated after over a year of development. I would at least wait until the changelog for that is published (and the compiler is tested) before making such a drastic change.

Is there an announced timeline on this? I'm generally in favor of the philosophy of this PR, but if there are about to be major compiler changes I'm sympathetic to that argument too.

lowercase-donkey:
I think that fixed the merge issues.

I'll spend time playing, but seems good so far over here.

pregAnon:
I will not be supporting this merge which, in turn, will lead to further code divergence down the line. 802+ passages to edit for no gain only compounds the feeling that I'm spending more time maintaining then creating content.

ObstacleCorpse:
> Is there an announced timeline on this?

No, nothing has been announced that I know of. Since development resumed, the compiled version is already at beta-2 though. Depending on how many beta versions he intends to do before the stable build, it may be sometime relatively soon.

pregAnon:
It's been discussed more and a conclusion has been reached. There would be no change on my end regardless if this is merged or not; the complications would come from large scale code migrations from my end to here.
However, there is a middle ground. If I updated the passages that lack a 'nobr' tag, then any remaining differences would be negligible. Now this is assuming there aren't something like 50+ passages that would need heavy alterations and that someone would be willing to provide a list of passages that would need updating.

All in all, much less of an inconvenience than all the descriptions being unconsolidated.

lowercase-donkey:
This sounds similar to the compromise I was thinking of, which is making sure that every page in vanilla has a nobr tag, and we could take or leave the global.

Is that it?

pregAnon:
Pretty much. It's not like they'd cancel each other out or anything like that. Though you never know with sugarcube.

lowercase-donkey:
Seems to keep everyone happy.  We just need to do it. What's the way forward for this commit then?

lowercase-donkey:
I have now reverted the removal of the nobr tags at the top of each page.

Current goal: every page of vanilla is nobr, but set locally.  Next step is checking files by hand to ensure this is actually the case.

lowercase-donkey:
Should be ready for merge.  

https://www.youtube.com/watch?v=abUGy_C7P84

freecitiesdev:
EXECUTE ORDER 66.

freecitiesdev:
See #940.

lowercase-donkey:
Yup, having a hell of a time figuring out what was going on with art widgets, but I thought I had the correct version with the final merge. Let me check up on it.

Edit: got it. a small update hadn't been merged in, but the main thing was that an old copy of the file had come back in the old location.

PR #919, opened 2017-08-02
Bugfixes

Folant:


freecitiesdev:
Thanks. These are bad; can't believe I didn't see a couple of them for how long I spent staring at that crap.

PR #920, opened 2017-08-03
Loot longSlave to make more widgets

lowercase-donkey:
Edit3: The focus of this branch, besides improving widgets, is bringing better descriptions to wardrobe, salon, and surgery.

Made pregnancy more visible in the corset section of the wardrobe.

Add a warning about extreme corsets if the slave is pregnant.

Then be generally out of hand.

Edit: From longSlave:
Make heelDescription for wardrobe to remind about heel status when doing footwear.
Make skinDescription for autoSalon.
Edit2: I can't stop, apparently.
Edit3: Major bugfix for shoulder/hip surgery.  Now functions like changing leg length: you can do the basic surgery once per slave, but need to get the advanced surgery to do unlimited changes.


lowercase-donkey:
I'll have to do another style pass at some point.  For now, I'm italicizing widget descriptions and leaving the old `surgery` descriptions in normal text.  However, some of those descriptions are simply inferior to the ones we already had in longSlave and fleshwidget, and I've pulled a few already where they are completely eclipsed.

lowercase-donkey:
Expand, contract.  Boom, bust.  Binge, purge (I guess).  After adding a lot of description widgets, I'm compressing the page back down again to make it better to use.

lowercase-donkey:
Yuck, I may need to split `VaginaDescription` in half in order to bring dicks and vag closer to each option in `surgery` for herms.  Ok to merge what I have so far though.

PR #933, opened 2017-08-05
Widget tweaking

lowercase-donkey:
Adjust accentDescription widget so it can properly appear while it needs to, while still being hidden where it needs to be hidden.

In longSlave, we don't need to hear about accents if a slave is not allowed to talk (That's how it is set up anyway. )

However, in surgery I think we're more interested in the raw biological possibility of the slave, and this is needed info when making major voice decisions.  Stronger weight to taking away a voice.

Minor: Standardize whitespace in the code here, so this looks like a bigger change than it is.


lowercase-donkey:
Promptly forget to start a new branch before committing, so I guess this is the "tweaking widgets" branch now.

PR #937, opened 2017-08-07
Tabs to Main

lowercase-donkey:
This adds tabs for each role to the Main screen.  Currently, it's very simple.

It does not "save" which tab player was on during page refresh.  Frustrating.  I'm sure there's a simple easy way to save what the last tab was between refreshes, but I'm not sure how to use .click() with a variable.

Ideally, I'd like to be able to take a tally of every slave role, perhaps show that number of slaves on the tab, and hide the tab if not in use.  Do we already tally somewhere?

Maybe there is a more compressed way to do what I did in code, so much is redundant.

Future: I'll also replace the "tabs" I added to the spa and such with these shiny ones.  I'd like to thank whoever made these?  Just need to get them to remember the last choice made, like the cheap tabs do.

Minor: indent changes to slave interact is all that happens below line 542, I was trying to learn what the code did.


freecitiesdev:
>Do we already tally somewhere?

No. The game tracks how many slaves are in each facility for obvious reasons, but not for slaves in the penthouse.

lowercase-donkey:
Sorry. I swear, sometimes I can rebase as one commit, sometimes it's commit vomit.

lowercase-donkey:
With this change, special slaves like HG always show at the top, reflecting DJs and Madams.  It's possible to make only HG appear here, and make a tab for BG and Recruiter if that is prefered.

Then Resting slaves are listed in a default tab.  Makes it soooo easy to find slackers and put them back to work.

PR #932, opened 2017-08-05
Shiny highlights for vector art

prndev:
Shiny clothing now optionally (enabled by default) has shiny highlights. This is probably as artsy as I will ever become on this project.

I am afraid to tell all of you that this is probably the last addition to this project from me for the immediate future. Please do not expect further content updates. I'll probably be around for bug-fixing, though. It is a pleasure to work with you.

PR #940, opened 2017-08-07
Revert "Switch to the new `nobr` global; remove nobr tags where they appear"

freecitiesdev:
Reverts Free-Cities/Free-Cities#918. That didn't take long.

The compiler doesn't like something about the art widgets when this is applied. It seems to revert recent art changes, losing the highlights added with #932 for example.


lowercase-donkey:
Yup, having a hell of a time figuring out what was going on with art widgets, but I thought I had the correct version with the final merge.  Let me check up on it.

Issue #935, opened 2017-08-06
Vector art is trapping dicks under boobs


lowercase-donkey:
Poor [Unnur](https://github.com/Free-Cities/Free-Cities/files/1202774/free-cities-20170805-202242.zip).  As you can see, she is either f**king her own boob, or kind of wedging it under there for safe keeping.

Joking aside, if each boob is a seperate layer, it might be cool to have dick drawn in front of rightBoob.  If they're together though, it's not really a problem.

prndev:
I noticed this problem. Boobs are not only in the same layer, but actually drawn with a single path.

lowercase-donkey:
> However, in case of very large boobs, it attaches the dick on the bottom of them. :/

Exactly so.  Thought this might be the case.  You clearly did the best with what ya had, and thanks again for your hard work!

freecitiesdev:
The way this used to work was that the dick and boob .svgs would be drawn in a different order based on the size of the body parts in question, thus:
- For monster tits, large enough to obscure the dick completely, the dick .svgs wouldn't be used at all, for performance reasons.

prndev:
It seems, this was fixed already in 07d6368.

freecitiesdev:
I apologize, I posted this and then circled back around, created that fix, and didn't update this issue.

There's still a (very minor) limitation with the fix: huge nipple piercings. Certain combinations of dick, boob, and piercings used to be displayed in the order boobs, dick, piercings, so that the chain between the nipples would be over the dick. Would it be possible to split nipple piercings and boobs so that the piercings can be done last?

prndev:
Yes. I dearly hope I got every special case now. :)

PR #949, opened 2017-08-08
Free willy from boobs with heavy piercings

prndev:
Splitup as requested in #935.


freecitiesdev:
>free willy

I'm glad I'm a part of this project.

PR #951, opened 2017-08-09
Bugfixes & art tweaks

freecitiesdev:
Reverts the back-and-forth on vector divs. No longer necessary.


racerand:
You have a lot of copied code here, is it really necessary??

freecitiesdev:
divs are a pain in the ass for someone with little to no experience. This is the least-worst way I've been able to get them to work reliably. I'm always open to suggestions, if you've got any.

racerand:
It's not possible to make some kind of function doing exactly these lines?? I haven't done much with javascript, but this seems like a thing that could be done in som kind of function, a widget maybe??

PR #953, opened 2017-08-09
Cleaner branch of the "Main" tabs

lowercase-donkey:
Contains what I was trying to do in #937, now with less messy commit problems.  As before:

Uses the 'weekly reports' style tabs on the `main` screen.  Currently, in an attempt to reflect the style of the Club and other buildings, "special" slaves are listed more statically, then the tabs appear below, with a tab for each common assignment (whoring, servant, etc.)  The default tab is "resting" slaves, encouraging players to check on their health, and consider reassigning them if their task just completed.  

Hopefully this helps performance as well, especially with art enabled.

All the old sorting works as well, allowing players to display only whores, but sorted by devotion.  (It may be possible to depreciate sorting by occupation after this.)

### DIFF
--- a/src/uncategorized/main.tw
+++ b/src/uncategorized/main.tw
@@ -172,7 +172,178 @@ __''MAIN MENU''__&nbsp;&nbsp;&nbsp;&nbsp;//[[Summary Options]]//
 <<if $positionMainLinks >= 0>>
 	<<MainLinks>>
 <</if>>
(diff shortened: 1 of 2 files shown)


lowercase-donkey:
I'm wide open to suggestions here, this touches something we all see a lot of and no doubt use a bit differently.  Should "special" slaves like HG/BG not be pinned and just have their own tab?

PR #955, opened 2017-08-10
Tabs with overview

lowercase-donkey:
Edit:  Goal is to add tabs to main screen, making it easier to find slaves that are on each task, especially those that are resting and need to get back to work.  Hopefully, this will ease the performance load as well, since typically only one role at a time is shown.  

There is also a tab for "all" the main slaves, and the entire tab thing can be disabled with the same setting as the econ tabs.

This also allowed me to add wiki entries to each role/tab to help new players, and veterans can turn this off with a new toggle in options.

Tested keyboard shortcuts for HG and such, they all still work.

This variant makes an "overview" tab for the truly important slaves such as HG, instead of pinning them above the tabs completely.

I would like to be able to add:
-Keyboard shortcuts to cycle tabs (existing shortcuts still work)
-Tab memory for when the page refreshes
-The ability to display the stat summary for the slave in personal training, if any.


lowercase-donkey:
One side effect of moving BG's italic events is that it also moves a little description snippet closer to her.

lowercase-donkey:
If this goes through, I suggest removing two things:
Sort by assignment.
"Sort options are SHOWN on the main menu. Hide"

The first is replaced by the tabs.

The second saves white space, but since it no longer appears on the first tab, I don't think it's a choice we need to force on players anymore.  It will appear when useful, and be hidden when not.

lowercase-donkey:
I can add a toggle for showing the wiki entries above each job, but I think it's a great way to remind players "what does this do, exactly."

I could possibly even do widgets for the text blurb, to make updates easier.

freecitiesdev:
What's the status between this and #967?

lowercase-donkey:
I'm ready to merge here.  I'd rather see the limited tabs stick around, and features added as we can.  I don't see a reason to remove tabs completely, especially when we have a toggle already for players that don't like them.

Frodfurgurn1991:
I believe this breaks the game for people who don't use the tabs (disabled in the options). Now all the special slaves are missing...

lowercase-donkey:
I'm working on it, sorry about that.

PR #962, opened 2017-08-11
Bits and pieces

Frodfurgurn1991:
Some simplifications and cleanups.

The biggest change would be the reworking of the facility arrays.  It used to be that they contained `{ID: ..., Index: ...}` objects, where `ID` is the ID of a slave and `Index` is its index in the `$slaves` array.  The was probably done for performance reasons, so as not to have to iterate over all the arrays just to find a slave, but it made the code complex because the `Index` could become "desynced" with the `$slaves` array (if a slave was somehow removed) and there had to be code to detect and fix that for each facility.

What I've done is as follows: the facility arrays now only contain IDs, plain integers.  In order to find the corresponding slaves, we build a mapping between IDs and their indices (in the `$slaves` array).  This is done by `<<SlaveSort>>` (since it also uses that mapping to do the sorting, for facility arrays) and is then saved in `$slaveIndices`.  Right now we rebuild this mapping in every `<<SlaveSort>>` done.  In the future we could build it once each week.

I admit I did not test thoroughly.  In what limited testing I did I noticed neither errors nor performance problems.  Please test yourself though.


Frodfurgurn1991:
After some testing, it appears that there are bugs.  **Don't merge yet!**

Frodfurgurn1991:
Ok, I believe that I fixed it, check it out (the comment about not testing much still applies).

src/uncategorized/penthouseReport.tw
```diff
@@ -11,21 +11,19 @@
...
 		<<include "Full Report">>
 		<br><br>
+
 		<<if ($slaves[$i].assignment == "be your Head Girl") && ($HGSuiteSlaves > 0)>>
-			<<set _iTemp = $i, $i = $HGSuiteiIDs[0].Index, _ID = $HGSuiteiIDs[0].ID>>
-			<<if $i >= _SL || _ID != $slaves[$i].ID>>
-				/% Slaves.ID and $HGSuiteiIDs.ID's don't match-up so let's find her %/
-				<<for $i = 0; $i < _SL; $i++>>
+			/% We found the Head Girl, now let's find her slave %/
+			<<set _ID = $HGSuiteiIDs[0].ID, _iTemp = $i>>
```
kopareigns: You need to remove the .ID here. On a related note, $HGSuiteiIDs does not seem to be added to BackwardsCompatibility.tw.

Also, in the very next line, why bother keeping the for loop when you could use $slaveIndices?

src/utility/miscWidgets.tw
```diff
@@ -1454,7 +1454,9 @@ This experience
 		<</if>>
 	<</switch>>
 <<else>>
-	<<set $args[0] = $args[0].sort(function(a, b){return a.Index-b.Index})>>
+	<<set $slaveIndices = {}>>
```
kopareigns: Should this not be initialized as an array?

kopareigns:
I have a couple comments/concerns about $slaveIndices.

1. This will be a very large array. If one is importing slaves via NG+, it will immediately have a length of *at least* 50,000, although all but a few entries will be empty. To my admittedly untrained eye, this seems like it might just sacrifice any minor performance gains for a significantly larger memory requirement. I'm not knowledgeable enough to determine whether this would be an overall improvement or not.
2. Why bother rebuilding $slaveIndices every time SlaveSort is called from a page other than "main?" It makes more sense to rebuild it *only* when called from "main" since that's the only time the $slaves array is sorted. In fact, as it currently stands, it will lead to errors if the slaves array is resorted, and the player continues to the next week without visiting a facility.
3. $slaveIndices is also easily modifiable when adding/removing slaves. While tedious to find all such cases, if correctly implemented when adding/removing slaves, this change could make it significantly easier (both on the eyes and processor) to find indices of slaves regardless of the current passage. I.e. it would broaden the scope of what you've done here.

src/uncategorized/penthouseReport.tw
Frodfurgurn1991: Thanks!

The reason was that I did not "build" `$slaveIndices` for that report and it may or may have not existed or be up-to-date.

src/utility/miscWidgets.tw
Frodfurgurn1991: No, this should indeed be a Javascript object.

Frodfurgurn1991:
> 1. This will be a very large array.
It will not in fact, as it is an object.  A JS object, will, in some cases that I don't remember off the top of my head act like a Hashmap.  Moreover, it only consumes enough memory to hold anything within it.

> 2. Why bother rebuilding `$slaveIndices` every time `<<SlaveSort>>` is called from a page other than "Main"?
It's just that, the places that use `$slaveIndices` right now are (mostly) the reports, and these always call `<<SlaveSort>>` at the top.  Moreover, `<<SlaveSort>>` itself as you can see uses `$slaveIndices`, so it's a natural place to build it as well.

> 3. `$slaveIndices` is also easily modifiable when adding/removing slaves.
Because that will be tedious both for the programmer (having to call some `<<RebuildSlaveIndices>>` or the like) and therefore also error-prone (not call it once and slaves may start duplicating or other horrors).

kopareigns:
> it is an object.

I did not know objects could be treated so much like arrays.

Unless I'm seriously misunderstanding something or missed something, my second point in 2 still holds. If the slaves array is sorted differently via main, and the player continues to the next week without visiting a facility, $slaveIndices will be out of date.

> Because that will be tedious and therefore also error-prone

I mean something just like adding the line
` <<set $slaveindices[$args[0].ID] = $slaves.length - 1>>`
in the AddSlave widget after the slave has been pushed to the slaves array. Adding a single slave shouldn't require an entire rebuilding of $slaveIndices. Similarly, when removing a slave isn't it a fairly simple matter to remove the relevant data from $slaveIndices? In fact, I don't see why this shouldn't be done anyways. Keeping old data seems more error-prone than removing at the first opportunity.

Or perhaps I'm still just too naive about javascript objects.

Frodfurgurn1991:
Your remarks did lead me to discover a bug, though it was unrelated.

To answer your question: given `$slaveIndices` is always built a-new in the places that use it, it is always up-to-date, even if `$slaves` is resorted.  Trying to identify all the places where `$slaves` is changed in any way would be hard.  And forcing the programmer to also change `$slaveIndices` to keep it up-to-date would be even worse.  It'll also be stupid: what I've written is not slow so trying to speed it up by caching would do nothing.

> "There are only two hard things in Computer Science: cache invalidation and naming things."
>
> -- Phil Karlton

kopareigns:
> $slaveIndices is always built a-new in the places that use it

I see this now and am apparently an idiot.

>forcing the programmer to also change $slaveIndices to keep it up-to-date would be even worse.

If you'll humor me a bit more, I'm still not understanding what would be bad about modifying $slaveIndices when adding a slave. It shouldn't take more than adding a single line of code in the widget `<<AddSlave>>`. Since this is the only way slaves are added to the $slaves array, no other changes would be required. It wouldn't alter the functionality of any pre-existing code, and anywhere (or at least in many locations) we're searching for the index of a slave, it could be replaced by looking at $slaveIndices.

A related side-musing is that $slaveIndices could also be used to check if we're about to add a slave with the same ID as another, and fix it if necessary. I've never actually encountered this issue, but it might be useful.

Frodfurgurn1991:
>  I'm still not understanding what would be bad about modifying `$slaveIndices` when adding a slave.

It wouldn't necessarily be bad.  However you'd then have to make absolutely sure that you keep `$slaveIndices` up-to-date (as you wouldn't be rebuilding it).  You say that only `<<AddSlave>>` modifies `$slaves` however that is not true, I found some passages that directly call `.deleteAt()`.

It would be nice to expand the usage of `$slaveIndices`, there are certainly uses for it.  This is something for the future.  I will look at it some more.

lowercase-donkey:
I know I'm like "the widget guy" but this one isn't mine.  I have no idea what people in this thread are talking about.

kopareigns:
Regarding your changes in 748bb02, I had set up the code like that in #849 to change the priorities of the Head Girl when training. I couldn't think of a good way to reduce the number of for loops since it's supposed to find all slaves with low health, then find all slaves with a paraphilia to remove, etc.

The goal was not to make it faster or look nice, it was to change training priorities. Without other major modifications, using only a single for loop would revert that change.

Frodfurgurn1991:
@kopareigns I understood the intent of the original code and I thought my rework kept the same semantics.  I had not thought about `$HGEnergy` though.  I'll think of a better solution (or just revert).

DrNoOne0:
<<AddSlave>> wasn't me, and I don't remember ever editing it.  I think it was created as part of a duplicate code clean-up by FCDev at some point.

Frodfurgurn1991:
Unless I'm understanding something wrong, it seems like [the Head Girl will first fix any behavioral flaws even if paraphilia are present](https://github.com/Free-Cities/Free-Cities/blob/master/src/uncategorized/hgApplication.tw#L179), even though [the code that finds slaves to fix checks for paraphilia first](https://github.com/Free-Cities/Free-Cities/blob/master/src/uncategorized/slaveAssignmentsReport.tw#L308)... Any reason for that?

lowercase-donkey:
Someone tweaked her behavior recently I know, with the idea that she should always scan the list for the worst problems and fix those first.  Might be tied to, but needing improvement?

kopareigns:
I don't think you're misunderstanding anything there. I see the same problem. When I changed the headgirl priorities up a bit I didn't mess with hgApplication.tw at all. Just changed which slaves were chosen and for what training. Although now that I look at it again I might see a small error with that too. I'll try and do some testing tonight if possible (time has been tight recently for that stuff).

Frodfurgurn1991:
@kopareigns I fixed the flaw in my current code (I think). What would you like to do about the rest of the problems?

Other than that, I think this PR can be merged... any more fixes can go in another one.

kopareigns:
From my reading that keeps the current priorities and fixes the small error I mentioned above.

Regarding the issue you caught in hgApplication.tw why not just replace [this line and the next 4](https://github.com/Frodfurgurn1991/Free-Cities/blob/40592785473da08e1fe45e38661988619686270e/src/uncategorized/hgApplication.tw#L177) with `<<set $activeSlave.sexualFlaw = "none">>`? I've only glanced at it briefly but I feel like that should be all that's needed. If you want to clean that file up a tiny bit more you could replace the flaw softening code with calls to the widgets `<<SoftenBehavioralFlaw>>` and `<<SoftenSexualFlaw>>`.

Another comment/question though. Is it possible to use only $HGTrainSlavesIDs for the calculations in slaveAssignmentsReport.tw? As in just find/replace _HGPossibleSlaves with $HGTrainSlavesIDs? Since you're just flattening/truncating _HGPossibleSlaves at the end and setting $HGTrainSlavesIDs to that.

Frodfurgurn1991:
Suppose a slave has a paraphilia but also a behavioral flaw.  Since both are registered as "flaws", even if the Head Girl is not to fix paraphilia, with the current code it will.

src/uncategorized/bodyModRulesAssistantSettings.tw
```diff
@@ -2120,7 +2120,7 @@ Leg tattoos: ''
 |
 <<link "Sacrament">>
 <<set $currentRule.legsTat = "sacrament">>
-<<RAChangeLegTattoosTattoos>>
+<<RAChangeLegTattoos>>
```
Folant: Other occurrences should be fixed too, probably.

kopareigns:
> she will actually fix paraphilia

Oh, I missed this before. Tunnel-vision I guess. I can think of two easy solutions
1. Add a check for $headGirlTrainsParaphilias
2. Make a new training type. Perhaps just make $HGtraining = "paraphilia" in that case and modify hgApplication.tw accordingly

Don't really have a preference.

> Is there a reason you'd want that?

Nothing huge, just seems like it might be a tiny memory save.

Frodfurgurn1991:
I said that she was fixing paraphilia when she shouldn't... That's not true actually, she was just mentioning that she was fixing paraphilia while in actuality it was the behavioral flaw that she was fixing.  Still, it was confusing I think.

freecitiesdev:
Can I get a review of this conflict, please?

Frodfurgurn1991:
Yeah, working on it

freecitiesdev:
Great. Merging, though please be vigilant for issues regarding the facility arrays.

If you're planning to make further contributions (and I hope you are, this looks good to my inexpert eye), please make a separate PR for something really radical like refactoring the way facility arrays work. That can make things easier if there's an undiscovered conflict down the road. It also allows more descriptive pull request titles than "Bits and Pieces." We've got nearly 1000 PRs merged and naming them descriptively makes life easier.

Frodfurgurn1991:
I know... I normally do smaller PRs but this time I just started working on things here and there and it got like this.

> really radical like refactoring the way facility arrays work.

It doesn't really seem that radical to me, though I guess with the trouble they've caused (duplications and stuff) caution is advised.

kopareigns:
@Frodfurgurn1991 just wanted to say thanks for being patient with my questions while I dragged this conversation through the mud. I blame it on little time and even less experience.

PR #964, opened 2017-08-13
New Recruiter Option

freecitiesbandit:
To mirror young migrant, a new option for the recruiter to focus on recent divorcees. They will be mature.


lowercase-donkey:
Always nice to see options rounded out.  Good way to cover mature slave acquisition.

PR #967, opened 2017-08-13
Adjustments

megas4o:
Small fixes, eco tabs removed until rework. Readjusted training, made no story sense that you would be as good at something after training for it for 15 weeks as something you have done your whole life


lowercase-donkey:
Ha, I just started using the tabs over here #955.  What needs to change, if I could ask?

megas4o:
It's designed to be standalone, but needs keyboard shortcuts to switch tabs, and needs to remember the last used tab. Basically a good idea, but it's for someplace else.

lowercase-donkey:
Well those happen to be problems I'm trying to solve right now.  Any idea how to get the click() to accept a variable?  Then remembering tabs would be pretty easy.

Pretty sure I could figure out keyboard shortcuts too.

megas4o:
it was a design oversight, so I'm redesigning the whole tabs system. If you can beat me to it, all the better.

lowercase-donkey:
I was just trying to understand what you did and copying what I could.  If it's ok with you, I'll leave the old tab code in the widget I'm using and if you can solve these problems at some point that would be pretty great.  I'm still using your toggle in `game options.`

megas4o:
changes to tabs have been rolled back, expect the new tabs in a fortnight

PR #970, opened 2017-08-14
p Peacekeepers Independence:  compare with numeric value

goldifrocks:


megas4o:
https://github.com/Free-Cities/Free-Cities/pull/967/files

goldifrocks:
ah, did not see that.  closing here then.

PR #975, opened 2017-08-16
fix for dickDescription and vaginaDescription

mikaelsson:
slave's description is incomplete when $saleDescription = 1

### DIFF
--- a/src/utility/descriptionWidgetsFlesh.tw
+++ b/src/utility/descriptionWidgetsFlesh.tw
@@ -1881,7 +1881,6 @@ $pronounCap's got a
 
 <<SlavePronouns $activeSlave>>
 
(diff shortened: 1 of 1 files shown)


lowercase-donkey:
Thanks, I wasn't sure where to cut it.

If those checks aren't happening in the widget though, shouldn't you make sure they happen when the widget is called?  They were there for a reason, whatever it was.

mikaelsson:
That two checks are just not needed. I followed changes in the commit you made (ca0f6ee). 
As I understand  `<<if ($showClothing == 1) && ($saleDescription == 0)>>` check is to describe clothing when slave is not on sale, it was starting at line 1342 and ending on line 1876, just before description of genitalia, including only part of the old widget.

PR #976, opened 2017-08-16
New FS acquisition events

freecitiesbandit:
A new acquisition event for every non revivalist FS - except chattel religionism.  14 new events!

Where events specified an age previously I tried to flip it, old to young/young to old, in the new event.

I'll probably get around to doing the revivalists as well, and I have an idea for chattel religionism. This was a lot of writing though, so I wouldn't expect that anytime soon.


freecitiesdev:
Great work.

PR #977, opened 2017-08-17
Colorful latex

prndev:
Latex comes in many colours. This update lets you select a outfit color via HTML5 color picker (tested in Firefox and Chrome).


lowercase-donkey:
Any chance of this coming to shoes?

prndev:
And for shoes! :smiley: (in 30f1d52)

lowercase-donkey:
Ayyyyy!  Not that I need to be color coordinated or anything, but god those blue heels were ugly.

prndev:
I probably should implement an "Arcology fashion colour" selector which overrides the default globally so you don't have to set it for each character individually.

lowercase-donkey:
Seems like a great idea; the RA might be a good place to add it.

PR #983, opened 2017-08-19
Fix missing things when tabs are disabled.

lowercase-donkey:


src/utility/descriptionWidgets.tw
```diff
@@ -16,6 +16,36 @@
...
+			<br>
+			<<if _BG > -1>>
+				''__@@.pink;<<SlaveFullName $Bodyguard>>@@__'' is serving as your bodyguard. <span id="manageBG"><strong><<link "Manage Bodyguard">><<goto "BG Select">><</link>></strong></span> @@.cyan;[B]@@
+				<<set $showOneSlave = "guard you">>
+			<<else>>
+				You have @@.red;not@@ selected a Bodyguard. <span id="manageBG"><strong><<link "Select one">><<goto "BG Select">><</link>></strong></span> @@.cyan;[B]@@
+			<</if>>		
+		<</if>>
+	<</if>>
 <span id="managePA"><strong><<link "Change plans">><<goto "Personal Attention Select">><</link>></strong></span> @@.cyan;[A]@@
```
Frodfurgurn1991: This should probably go above.

lowercase-donkey: Sorry pretty sick atm.  Thanks for checking on this.  If there's anything else I screwed up I'd be grateful if you go ahead and fix it.

pregAnon:
Would it be possible to get you to seperate the main tabs from the economy tabs so that choosing one doesn't disable the other?

lowercase-donkey:
Sure.  Let me talk about tabs for a sec though.

I added taaaaaabsish? to the brothel and spa and such, allowing players to cycle between "add a slave" and "subtract a slave."  I'd like to replace this with Proper Tabs, especially since we'll have keyboard shortcuts coming soon.  I don't want to do this until the "new" tabs have tab memory though, because my kludgy way of doing it has at last that.

My question is: how many places do we want tabs toggled?  How long do we support a legacy UI?

I didn't add a toggle at the time because I thought one was enough for people that wanted FC to look the "old" way.

Is there a reason people would have some tabs on and some tabs off?  How many degrees of choice do we allow as tabs continue to grow?

pregAnon:
Slave tabs definitely should not be tied to economy tabs, for starters. Secondly, the tabs used in main are visually unpleasant; they are much too large and furthermore hide certain elements, like using your fucktoys, behind certain tabs.

Ultimately my opinion is moot, however, since my concerns are overidden by those above me having outright rejected it for merge and they are very unlikely to be swayed.

lowercase-donkey:
Fucktoy scenes could go anywhere, technically.  And they aren't hidden.  If you want to fuck your fucktoys, go to the fucktoys tab.  If you want to fuck your bodyguard, go to the tab with the bodyguard.  Aside from a thematic connection, I changed it because they're much more visible there than at the end of a list that may be hundreds of slaves long.  Hopefully the scenes are more visible, not less.

As for tabs being large, I agree.  They're seeing a bit of a rework and hopefully that is adjusted.  Perhaps they could be sized to correspond to the arcology "rooms."

PR #996, opened 2017-08-21
HG flaw fixing and BC fixes

pregAnon:
She'll now do her job right. Also parentheses are your friends.

Also tossed in a clearer to kick mindbroken slaves out of the cellblock since they confused slave summary to the point of getting stuck in the facility.

And now more: Backwards Compatibility will now properly set the HG suite slave array and stop thinking it is running 0.6. Strongly advise moving to 0.11.0.0 to avoid trouble with 0.7 through 0.9.

### DIFF
--- a/src/uncategorized/BackwardsCompatibility.tw
+++ b/src/uncategorized/BackwardsCompatibility.tw
@@ -991,11 +991,11 @@ Setting missing global variables:
 	<<set $masterSuiteDecoration = "standard">>
 <</if>>
 
-<<if ndef $HGSuiteiIDs>>
-	<<set $HGSuiteiIDs = []>>
+<<if ndef $HGSuiteIDs>>
+	<<set $HGSuiteIDs = []>>
 <</if>>
-<<if $HGSuiteiIDs.length > 0 && typeof $HGSuiteiIDs[0] === 'object'>>
-	<<set $HGSuiteiIDs = $HGSuiteiIDs.map(function(a) { return a.ID; })>>
+<<if $HGSuiteIDs.length > 0 && typeof $HGSuiteIDs[0] === 'object'>>
+	<<set $HGSuiteIDs = $HGSuiteIDs.map(function(a) { return a.ID; })>>
 <</if>>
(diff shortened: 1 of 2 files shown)


src/uncategorized/BackwardsCompatibility.tw
```diff
@@ -991,11 +991,11 @@ Setting missing global variables:
 	<<set $masterSuiteDecoration = "standard">>
 <</if>>
 
-<<if ndef $HGSuiteiIDs>>
-	<<set $HGSuiteiIDs = []>>
+<<if ndef $HGSuiteIDs>>
+	<<set $HGSuiteIDs = []>>
```
kopareigns: This should in fact be `$HGSuiteiIDs`

pregAnon: Then we have a different problem. `<<set _ID = $HGSuiteiIDs[0], $i = $slaveIndices[_ID]>>` in penthouseReport.tw is failing to propery define $i and I have been dealing with reports all day. BC was having no effect on it, though it works on ng/ng+.

kopareigns: @Frodfurgurn1991 might have a better idea what the issue is. Any ideas I have are little more than grasping at straws. If I had to try something, I'd try putting `set $args[0] = ` in front of [filterInPlace here](https://github.com/Free-Cities/Free-Cities/blob/master/src/utility/miscWidgets.tw#L1459). But again, this is just grasping at straws.

Frodfurgurn1991: Does there happen to be a save with the problem that I can dissect?

pregAnon: Not currently. I also can't get the bug to occur. The way to cause it seems to be having a HG suite with slave in it from prior to the facility changes.
Now that said, everything else with the changes seem to be working fine. If it turns out to be isolated and not recreatable, I can easily circumvent it. But it is something to keep an eye out for.

PR #1001, opened 2017-08-23
RA Changes

Frodfurgurn1991:
I set out to "finish" what I started in #420, that is to also merge `setAssignment` and `assignFacility` before applying them so that you wouldn't get stuff like

> Foobar has been automatically assigned to work in the Brothel.
> Foobar has been automatically assigned to relax in the Spa.
> Foobar is now allowed to select her own assignments.

over and over again.

To make the above easier, I eliminated `assignFacility` and just use `setAssignment`.  This means that you don't have to write `<<set $currentRule.assignFacility = "foobar", $currentRule.setAssignment = "none">>` all the time.  It does mean that in places where you need to know if a slave would be set to an assignment or sent to a facility you now need to check against a list of those but that's literally just three places right now.

Also, I changed how inclusion and exclusion works for the rules.  It used to be that if a rule was setting applying or excluded facilities, only the slaves that were in facilities in general would be the ones excluded.  Like, if a rule said "I only want to deal with slaves in the Brothel", it would also apply for all the slaves in the Penthouse (because of `assignmentVisible`).  I found that weird and so I changed it.  In the example, only slaves in the Brothel would be considered at all.

The less impactful changes are as follows.

First of all, instead of having to write `<<RAChangeSave>>` and `<<RAChangeApply>>` given that these two (almost) always come together, there's now a macro that combines them, `<<RARuleModified>>`.

Secondly, where before the widgets that deal with assignments in the RA UI had a ton of code in the links, I now moved all of that in `<<RANormalizeAssignments>>`.  You just have to change the variable that the user actually tells you to change and let it handle all the collateral including updating any UI that needs updating.

This ended up being bigger than I originally thought.  The UI changes I have tested and seem to work ok.  The rest, **please test yourself**.

### DIFF
--- a/devTools/tweeGo/targets/sugarcube-2/userlib.js
+++ b/devTools/tweeGo/targets/sugarcube-2/userlib.js
@@ -316,222 +316,39 @@ window.ruleApplied = function(slave, ID) {
 	return slave.currentRules.includes(ID);
 };
 
-window.ruleAssignment = function(applyAssignment, assignment) {
-	if (!applyAssignment)
-		return false;
-	return applyAssignment.includes(assignment);
-};
+window.expandFacilityAssignments = function(facilityAssignments) {
+	var assignmentPairs = {
+		"serve in the club": "be the DJ",
+		"rest in the spa": "be the Attendant",
+		"work in the brothel": "be the Madam",
+		"work in the dairy": "be the Milkmaid",
+		"work as a servant": "be the Stewardess",
+		"get treatment in the clinic": "be the Nurse",
+		"live with your Head Girl": "be your Head Girl",
+		"serve in the master suite": "be your Concubine",
+		"learn in the schoolroom": "be the Schoolteacher",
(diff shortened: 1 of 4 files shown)


pregAnon:
I've gotten some reports about the RA fighting with the HG suite over diets and the Master Suite over living conditions. Think you can take a look at any possible causes?

Frodfurgurn1991:
@pregAnon if by fighting you mean that the HGsuite/MasterSuite sets it to something and then the RA sets it to something else then I'm afraid there's not much I can do about that, other than not letting the RA apply to those two facilities at all.  If you mean something different, please explain.

pregAnon:
I meant exactly that. Though if your HG is allowed to change her slave, that might be enough of a control shift to warrant disabling the RA on her, thoughts? No way that could be though of for the MS, however.

megas4o:
https://github.com/Free-Cities/Free-Cities/issues/1005

PR #1022, opened 2017-09-13
Version update

Weresmilodon:
Changes $ver to 0.10.7 and $releaseID to 1020.
The new upload to the blog also still say 0.10.6.


pregAnon:
0.10.7 will cause problems with backwards compatibility. Same deal as 0.10.6; the code sees the "0.7" and starts changing things it shouldn't.

Weresmilodon:
I assumed this change was simply forgotten. This was made only because the new post on the blog says 0.10.7, while the number ingame does not.

freecitiesdev:
This should have been included; not sure why the release PR didn't.

> 0.10.7 will cause problems with backwards compatibility. Same deal as 0.10.6; the code sees the "0.7" and starts changing things it shouldn't.

Damn, I thought we had a fix for this. I'll do that.

PR #1027, opened 2017-09-15
(Read Before Merging) TweeGo, SugarCube, and Associated Updates

ObstacleCorpse:
TweeGo was finally updated to v1.0.0 last month. After seeing the v1.0.1 bugfix, I was waiting to see if a v1.0.2 ever emerged. It has not, so I'm assuming that things are stable now.

TweeGo's documentation can be found here:
https://www.motoslave.net/tweego/docs/

There are several changes, but the one with the largest impact on FC will likely be the removal of the StoryIncludes passage. You no longer have to specify a start.tw passage for TweeGo, or maintain a list of *.tw files to be compiled. Just supply the source's root directory (./src/) to TweeGo, and it will find everything.

I've updated the compile scripts for this new method, and everything is working on my end. However, I have not tested the resulting game extensively. I have also only tested compilation on Windows, and have not tested the additional bash scripts in the root directory. Anything that touches the StoryIncludes.tw file will need to be changed, and anything that touches the start.tw file might need to be changed. I have archived my repository here if you'd like to test everything before merging:
https://mega.nz/#!uYZViZCb!J95Syl9Ry86M9-iRjOrxQDaLrEPele1VIWg-5RU-zEc

Additionally, SugarCube 2 has been updated to v2.20.0. Release notes:
https://www.motoslave.net/sugarcube/2/releases.php


freecitiesdev:
I'm merging this to force it to the attention of anyone using this repository, and thus to get more people testing it. I will do some serious playtesting before releasing a public build based on this. Thank you as always.

Frodfurgurn1991:
This removes 'src/config/start.tw.proto'.  It should only be renaming it to 'src/config/start.tw' since the Start (and StoryTitle) passage are needed by tweego... at least that's what the error messages tell me.  Moreover, it should also remove it from gitignore since it's now used as-is.

Issue #1039, opened 2017-09-26
Penis enhancement failure


eldestdawn:
The dairy is not properly applying penis enhancement to slaves. All slaves in the dairy that have a penis and are on penis enhancement drugs don't show any results on the weekly report. They also don't change in their character sheet on the following week. Other enhancement drugs work, it's just the penis enhancement that doesn't. Slaves assigned to be Milked in the Penthouse however are not affected.

prndev:
By looking at the code, I can tell you the dairy grows a slave's dick only if all of these criteria are fulfilled:
  * feeders are installed and active
  * FS slimness enthusiasm's "keep cows slim" is not activated
  * slave has balls
  * dick size is less than 10 (50cm)
  * random trigger

I checked the game's actual behaviour (42e5e78 with patches from #1028) and dicks were grown appropriately. Bear in mind this dick growth is not mentioned textually. Not in the weekly report nor anywhere else.

Usual dick+ drugs are applied independently. Their effects are mentioned in the weekly report.

Fun fact: While examining the code, I discovered an unrelated bug. Balls are grown only if the dick is small.

eldestdawn:
Considering I have multiple cows with cocks from early on in my game and all of them have 10 cm penises...I think there's something wrong with my dairy. Currently on week 212 so they should be max penis size by now. One of the cows has been with me for 137 weeks. Here is the save, just look in the dairy. 

[free-cities-20170926-142810.zip](https://github.com/Free-Cities/Free-Cities/files/1335062/free-cities-20170926-142810.zip)

I want to know if there is a way to fix this or if I have to start a completely new game.

prndev:
Your dairy does not feature the "intubators" (automated feeders).

eldestdawn:
While the Intubators aren't active, the cows were still on penis enhancement for a while but the weekly report never mentioned them receiving them. I had rule 10 set up to get their penis size to porn star (it's targeting any slave in the dairy or being milked). The weekly report didn't even mention the failed injections "Despite being dosed with penis enhancement hormones, XXXX's cock doesn't grow" or something like that.

prndev:
I now understand that you are referring to the Rules Assistant. I did not notice that earlier. Indeed, I found the Rules Assistant to be broken in this regard. Please also keep in mind that fertility drugs override enhancement drugs.

PR #1043, opened 2017-09-29
RA "injections" vs "enhancement"

prndev:
Rules Assistant no longer applies "penis injections" instead of "penis enhancement". Closing #1039.


lowercase-donkey:
While you're in there, you might want to update "testicle injections" to "testicle enhancement" as well.

PR #1044, opened 2017-10-01
Attempt to bring in line with waist changes (not sure if correct)

mweggen:


pregAnon:
I opted for `$slaves[$i].waist < -95` for the first part since sub 40 can be achieved with extensive corsetting and the overall focus seems to be surgical modifications vs lack of said mods. Same thing should also be present in clubReport.tw as well as buried somewhere in assayWidgets.tw.

mweggen:
Updated it

PR #1046, opened 2017-10-02
Take advantage of the new TweeGo

Frodfurgurn1991:
We can now use '.js' for JavaScript!  Isn't that great?  It means we now get automatic syntax highlight!  Same for CSS.  I also added back the 'start.tw' file (removing it from the gitignore as well) since it's needed by TweeGo.


Frodfurgurn1991:
Rebased

PR #1051, opened 2017-10-12
Assorted Fixes

kopareigns:


kopareigns:
Summary so far:

- fixed a misnamed variable
- updated old fetishStrength values
- fixed up the $RapeableIDs array and the passages that use it
- fixed up RETS in the case where the player clicks the slave's name during the event

PR #1055, opened 2017-10-22
Add dye colors to vector art

lowercase-donkey:
Things like Camouflage and Tiger Stripes I just went with olive drab and yellow.  Better than nought.


lowercase-donkey:
@freecitiesdev you mentioned a while ago you might be willing to put an "original skin" variable in.  It would be handy here, since vector art doesn't know what to display with an un-dyed slave.

Right now that slave simply has "natural" skin, which is a nice catch for text but is hard for vector art.  For now, #bde460c shows them as tanned.

PR #1058, opened 2017-10-27
Fix clothing "uncomfortable straps" for flat chests

prndev:
Clothing "uncomfortable straps" no longer shows straps for boobs on slaves with flat chests.


prndev:
No, wait, hold it. I messed something up.

prndev:
Okay, found the mistake.

PR #1074, opened 2017-11-08
Headings and Surgery tweaks

lowercase-donkey:
Expand the use of headers to some of the basic rooms.

Looks like these were first added in 8e68b3c but used very rarely.

This way it's not just used in bodyMod.

Edit: sorry for the extra things that got clustered in here, not sure how I did that.


lowercase-donkey:
Merge the surgery changes myself because it was going to be messy.

Made a mess anyway.

Fixed mess.

lowercase-donkey:
"The four rooms" now matches a kind of mental style guide, which I'd be happy to write out or adjust to feedback.  Ideally though it's intuitive, and by using headings it will be even easier to adjust later.

Ideally though:

**Heading**
_Italic description of the area of the slave being addressed in the heading_
---    Indented options.  Where needed, describe area or action in white text.  [Option](). _Warnings afterward in italics._
--- ---   Optionally, double indent below that and present [option]() | [option]()

PR #1078, opened 2017-11-20
Fix syntax error in autosurgery eye logic

Scriptifaber:
The `>>` syntax was missing from the end of the line, causing errors when applying autosurgery at the end of a week.


devartie:
This fix is included in #1114 as part of a major cleanup.

PR #1084, opened 2017-12-01
Remove odd flat bonus to beauty for being white or white-ish, no matt…

lowercase-donkey:
…er what policies were chosen.

Via reddit: https://www.reddit.com/r/freecitiesgame/comments/7gs8nf/white_slaves_get_an_unconditional_boost_to_their/

I agree with the reddit post, not because PC, but because it makes no sense in FC lore, where you might be opening an arcology anywhere on the planet, and might choose to fetishize just about any race.  There is currently code for supremacist/subjugationist beauty bonuses, so I think this may have been a legacy thing.

If we decide to keep this, I feel it should be made more transparent to the player as well.


Boney-M:
From the Encyclopedia:

>white slaves enjoy a minor bonus to beauty, modeling the near-universal reach of western standards of beauty.

lowercase-donkey:
So it is deliberate, then.

PR #1087, opened 2017-12-04
Improve tracking and display for slave costs

lowercase-donkey:
Add three variables for slaves: their initial cost, and the expense and income they accrue as time passes.

Display this info in ``Upkeep``.  

Add ``Upkeep``'s info to ``slaveInteract``

Widgetize when necessary.


lowercase-donkey:
Going to work on this a bit more later, adding some other cases when slaves come on board through events.

Hoping to be able to show this new data at point of sale too.

lowercase-donkey:
Commit should now keep track of how much slaves cost (if anything) when acquired, and how much they have gained in the past week, and how much they have gained and lost over all time.

For ease of all of us, created the widget `slavePaid` that should replace updating cash directly in slave events, as it will update slave records at the same time.

So `<<$cash += 100>>' becomes '<<slavePaid $activeSlave 100>>``

It also correctly handles negative values, updating the slave expense variable instead of their income.

Should keep reddit happy.

pregAnon:
Advice:
I question how many people want to slog through all the slave costs when trying to interact with them. I suggest setting up a replace to swap it in for those who want to see it rather than have it always there.
Secondly, why call `<<PerSlaveCostCalculation>>` and do the update in persBusiness instead of when the weekly costs get processed? It would be more efficient to do it during the slave costs calc loop, you know? One less full slave array loop to go through.

lowercase-donkey:
"I question how many people want to slog through all the slave costs when trying to interact with them. I suggest setting up a replace to swap it in for those who want to see it rather than have it always there."

I can add a toggle, sure.

"Secondly, there has to be a better place to call <<PerSlaveCostCalculation>> and do the update than in persBusiness. Is there no other full slave array for loop that it could be added into?"

Hahahaha, that one's totally not my fault.  The old line there was ``<<set $cash -= $costs>>`` and it took me forever to find.  That's what causes the cost calculation to actually fire ("payday" if you will."

I assume it's done there because ``persBuisness`` only fires at the end of the week.  If you have a spot that makes more sense, we could try to move it.  I'm not clear on all the pages that fire besides personal business.  longTermeffects or assaywidgets maybe.

pregAnon:
When it gets merged I'll consider things. I'll have enough conflicts to handle to force my attention to it, assuming I haven't lost patience and rebalanced the entire economy.

FCdev's absence does bad things when it comes to me and code divergence.

pregAnon:
nextWeek.tw has a viable for loop in it and is the last step before a new week begins. It would seem to be the ideal place.

lowercase-donkey:
Thanks!

lowercase-donkey:
That should be it.  I'll probably tinker a bit more for a while, but the books balance most of the time.

The only things left I can think of would be adding slave purchases, rare income such as from from the gambling event, arcology cladding, (basically "one-off" events.)  I'm reluctant to touch more pages before getting this merged though.

Also maybe chasing down a few smaller variances, but they're only throwing off the totals by tiny percents. I'm very pleased with it.

lowercase-donkey:
Yiiiiiiiiiisssssssssssss!  Finally found the thing I wasn't recording, and it was slaves generating milk on the side when they had a non-cow job.  Sheesh.

PR #1091, opened 2017-12-07
Bugfixes

kopareigns:
Missing angle brackets, off by one error, changed == into >=, and changed $ into _


kopareigns:
While not really a bugfix, the Manage Headgirl screen was not clear that both $headGirlTrainsFlaws and $headGirlSoftensFlaws had to be toggled to enable softening of flaws. This latest change allows only $headGirlSoftensFlaws to be toggled to soften flaws.

